PR #240, opened 2016-05-27
Release0 3 1

PRemmen:
This is release 0.3.1 with updated documentation and READMEs, extended example for better understanding how to set materials and PyPI (and pip) support of teaser.


MichaMans:
pytests are running. Examples are working, Docu looks good. I'll merge it.

PR #248, opened 2016-06-01
fix bug and update documentation

PRemmen:
this should fix wrong function call


MichaMans:
Should solve the issue, i'll merge

PR #249, opened 2016-06-01
Change if statement for rooftops

PRemmen:
This is just a minor bugfix, as it will only occur if no rooftop is defined


MichaMans:
this should work, i'll merge this and #248 for updating the version numbers

Issue #260, opened 2016-06-21
TEASER is incompatible with Unix-like operating systems


yfuerst:
TEASER declares paths with double backslashes ("\\"), which is Windows specific. Unix-like systems use forward slashes ("/") and need no escape sequence. I am therefore unable to execute TEASER on my Linux machine.
`os` is clever and works out which slashes you need for your OS. A better way, than providing the slashes yourself, would be the use of `os.path.join`. Other solutions could be the use of raw strings or forward slashes, which will be substituted by Python with the correct slashes, according to the OS.

PRemmen:
@yfuerst  thanks for raising that issue. We are aware of this bug ( #181 and #214 ) and hope to resolve that issue soon. I can't promise exactly when, but we will keep you in touch.

MichaMans:
This should be solved with PR #263

PRemmen:
@yfuerst the latest release 0.3.6 should solve this issue. Please update you teaser distribution by using pip. If you find any problems, please reopen this issue :-)

yfuerst:
I just pulled from git and the issue seems to be solved. Thanks for the fast action.

Issue #214, opened 2016-04-12
TEASER does not like paths that use "/" and "\\"


marcusfuchs:
Problem occured in master:

`Project.export_record(...path=...)` fails for `path=D:/Python/uesgraphs\workspace\osm\TEASER` but works for `D:\Python\uesgraphs\workspace\osm\TEASER`.

MichaMans:
This should be solved with PR #263

Issue #181, opened 2016-03-23
Revise the os path interactions


MichaMans:
We should revise the os path interactions! There is a function in the utilis which we should use everywhere. In addition we should revise these utilis path functions to be more stable, especially if we want to use same linux machines.
So:
- revise utilis path functions
- revise all functions with path arguments

MichaMans:
This should be solved with PR #263

PR #262, opened 2016-06-22
Issue181 ospaths

MichaMans:
Revised paths and os.path functions, in addition i added numpy and scipy to the setup file 

should solves #181 , #260 , #214


PRemmen:
will be merged with release PR

PR #264, opened 2016-06-28
Issue242 doc

yGoerguelue:


PRemmen:
Does not seem to break to gui, so I can merge this,

PR #276, opened 2016-07-20
allows defining weather file path and weather file header

### LINKED ISSUE #256: Customizing weather_file and weather_file_header
With one of our releases, we deleted customizing weather file path and weather file header. Infrastructure is still entirely there, it's just not set in the template.

mlauster:
closes #256


PRemmen:
You need to update some version numbers (see https://github.com/RWTH-EBC/TEASER/wiki/Workflow#bugfixes) after that this is ready to merge.

PR #289, opened 2016-08-02
Issue93 create building from xml

yGoerguelue:


MichaMans:
@yGoerguelue could you check if your update method works on every building in the project building list, because in a first view it looks like it always updates the last building in the list and not the actual building selecetd. (This is all fine for the function that we want to implement here, but we can also use the update building function to change, e.g. the type of building already in the project)

yGoerguelue:
@MichaMans The update method works on every building in the project building list. My Problem was that i couldn't change the type of a Building. That's why I generated a new building with the updated values and if a new building is generated it automatically added to the end of the project building list.  If you toke a deep look at the function click_update_building in the controller (screenshot) you can see that i pop the last and the selected building off  and add the updated building into the right place. 


click_update_building

MichaMans:
ok looks good! i'll merge it!

PR #293, opened 2016-08-10
switches from MSL 3.2.1 to 3.2.2

### LINKED ISSUE #291: update "uses"-statements to MSL 3.2.2
In aixlib_output and annex60_output, we have an attribute describing with libraries in which versions are in use for our projects. With Dymola 2017, we have version 3.2.2 of MSL. We need to thus our "uses" attribute, otherwise Dymola asks for each run if the package should be updated to MSL 3.2.2.

mlauster:
this closes #291


MichaMans:
sorry didn't saw my assignement. Now i'll merge. Changing two digits in two files should not break TEASER :)

Issue #298, opened 2016-08-30
ThermalZone: inner wall attributes missing in set_calc_default


PRemmen:
reported by @bruentjen 

The inner wall attributes (e.g. area_iw) are missing in the ThermalZone.set_calc_default() function, this causes wrong output values if a building is calculted twice in one project (e.g. because of retrofit).

MichaMans:
merged with #299

PR #299, opened 2016-08-30
Issue298 set default bug

PRemmen:
Unit Tests are running in python 3.5 and 2.7, should solve #298


MichaMans:
Pytests are running. Looks good to me. I'll merge and push a new release

Issue #296, opened 2016-08-10
Linux


jack-ew:
If you import teaser with:
`from teaser.project import Project`

you get the error about write permissions:
`[Errno 13] Permission denied: '/usr/lib/python3.5/site-packages/teaser/OutputData'`

In linux you don't have this write permissions for your normal Systempath. I think the simplest solution would be a function that asks for a different path

pema1:
Having an output directory inside the package structure is not really nice. Perhaps Python can be made using the user's home directory for outputs as a default. That should work on any platform, hopefully.

@jack-ew maybe you can use a _virtualenv_ in your home path to install and use teaser for the time being.

PRemmen:
@MichaMans is this ready to merge for new release (see other 4 bug pull requests)

MichaMans:
I haven't tested it yet, so i would suggest to test it first. The problem is, since this is a pip related error we need to install this branch with pip and then test the functionality. Do you have any suggestions how to do so?

PRemmen:
@jack-ew we solved this issue with latest version (v0.3.9) can you verify that?

PRemmen:
closed by #311

PR #308, opened 2016-09-21
Issue307 pyxb

PRemmen:


PRemmen:
@MichaMans is it possible to integrate issue #296 into release 0.3.9

if not I will merge the 4 pull request and then you can review this PR and we publish next release.

MichaMans:
Why aren't we merging this branch #308 into the 0.3.9 release and then publish the new release? #296 can be integerated as well, if you say its a good implementation and if it works for you? Testing if this really solves the Linux issue can wait until the end of the week. for now we can handle #296 as an improvement for the TEASER output in general. Until we can confirm that the problem is solved we can let the issue stay open. Do you agree?

PRemmen:
okay agree

PR #323, opened 2016-10-13
Polishes all icons, e.g. making them transparent

mlauster:
Nice new icons and reworked names and descriptions


PRemmen:
All model exports can be loaded and started in dymola 2017 so I'll merge it.

PR #325, opened 2016-10-14
Issue324 namespace

PRemmen:
This should enable Python objects to be pickled


MichaMans:
It wasn't the code, it was the old file which throws the xml error. So all good with that. I'll merge

PR #333, opened 2016-10-28
fix bug for warnings in gui before gui is initialized

PRemmen:


MichaMans:
This seems to solve the problem. I'll merge and after #336 i'll release a new version

PR #339, opened 2016-11-15
Issue337 alphas

### LINKED ISSUE #337: wrong values for alphas
For `AixLib` exports, the indoor coefficient of heat transfer for interior walls (`alphaiwi`) is always set to zero. Oddly, it seems to work for `Annex60`. 
For all indoor alphas, the value is set to convective plus radiative (so e.g. 7.7 instead of 2.7), although even in the `AixLib` models, we handle radiative and convective heat transfer in separate circuits.
@PRemmen, can you help to track down the bug, it seems to be quite urgent as this affects all models and leads to wrong results.

mlauster:
This is to fix various bugs in templates and template-related functions

I mistakenly included one commit that is not related to that topic when in rebased my branch as I mistaknely started from `development`. As this is only cleaning up GUI layout, I hope it is okay to merge it anyway

This closes #337 

@PRemmen, I tried to cover all necessary steps (increasing version number etc.), could you please check that and run the regression tests?


PRemmen:
To comply with workflow i reset `mainui.py` to master and fixed a bug in the readable output template as revealed by unit tests.

Now unit tests are running and from my perspective these changes fix the mentioned bug + something extra. I can only comment on the programm and workflow and can't verify parameters of Annex, at some point we would need some documentation how and why the parameters from aixlib and annex differ

PR #348, opened 2016-12-02
Issue343 hotfixes

### LINKED ISSUE #343: Collect hotfixes while debugging
While debugging the all-new models for https://github.com/RWTH-EBC/AixLib/issues/260, I occasionally find bugs in the current AixLib-models. So far, all of them were related to parameters of TEASER export. This issue is to collect and fix these bugs.

So discovered bugs:
- [x] The exterior wall area `Ao` contains the window area `Aw`. In the current export, `Ao` is equal to the exterior wall area without window area
- [x] `alphaowo` must not contain radiative heat transfer coefficient, a coefficient of 5 is added automatically in the AixLib model
- [x] `awin` should not be zero, in the new models 0.03 is set as default value within the thermal zone model. Let`s take this value as well in this bugfix.
- [x] `epsw` is always close to zero, what does not seem to be reasonable. In all standards and intrinsically when linearizing radiation, a value of 0.9 is taken.

I will do some more testing and then make a PR to fix these issues.

mlauster:
Fixes bugs that came up while debugging the all-new models. This fix should fix all found bugs, the all-new models deliver almost the same results as the current models with this fix. All related deviations can be explained by modeling differences.
This closes #343


PRemmen:
It might be worth it implementing another bugfix for heatload in this issue? see https://github.com/RWTH-EBC/TEASER/commit/4b7da5ec1e90ad4c0b6554a954349ce524c95fbf

in addition with this relase in the master i will add How to cite section

mlauster:
Good idea! How is taking care of this? :) @PRemmen, do you have a verified fix for this?

MichaMans:
I added a few outstanding things. @PRemmen could you have a look again.

PRemmen:
This is ready to merge, models `archetype.py` and `singlebuilding.py` simulate using Dymola 2017 and AixLib v0.3.2.

Unittest 49/49 for py35 and py27

Issue #349, opened 2016-12-06
Release 0.5


PRemmen:
For release 0.5 I would like to merge some big issues into one branch:
#347, done 
#341, done

PRemmen:
#351

PRemmen:
#342

PRemmen:
#344

PRemmen:
#352

PRemmen:
#353

PRemmen:
#362

MichaMans:
#361 merged into Release0.5

Issue #347, opened 2016-12-01
Restructure ThermalZone class (especially calculation)


PRemmen:
Currently the ThermalZone class is rather large holds lots of functions and attributes that are not needed for standard case (e.g. all calculation attributes for four-element model). The idea is to separate the calculation of the parameters from the ThermalZone by creating four new classes (one for each model) and saving functions and attributes related to the calculation method in that class. This class will then be attached to the ThermalZone with an attribute.

We could also extend the lists for outer, inner walls to alls possible type (e.g. list for outer_walls, ground_floors and rooftops).

**The API of thermal zone, building, and project class should not change**

PRemmen:
It might make sense to also restructure the building class in the same manner (to have AixLib and Annex classes, because we have specific attributes (e.g. pathes for internal gains). Any objections that this will be done in the same branch?

mlauster:
No, go for it

mlauster:
Another item to solve within this branch: We currently follow the definition of attributes according to VDI 6007. Unfortunately, these definitions are in some parts specific to the VDI calculations. E.g. `r_rest` contains the outdoor coefficient of radiative and convective heat transfer `alpha_comb_outer_ow`. For the Annex and AixLib models, we need the resistance (let's name it r_rem) separate to heat transfer coefficients. Thus, we substract `alpha_comb_outer_ow` from `r_rest` in the templates. We should find an own definition and descriptive names within TEASER that represent the smallest parts we need and avoid post-processing in templates. Regarding resistances and capacitances, I suggest to have two lists for resistances and capacitances each and one resistance separate that is the one between exterior surface and capacitance `r_remaining`. This one ensures a valid U-value.

PRemmen:
done in #349, stays open until release

PRemmen:
done and merge in #365 stays open until released

PR #350, opened 2016-12-09
Issue318 aix lib export

mlauster:
This updates the export to the all-new Modelica AixLib models
This closes 318


teaser/data/output/modelicatemplate/AixLib/AixLib_Multizone
```diff
@@ -0,0 +1,377 @@
...
+      %endfor
+      },
+% if bldg.central_ahu:
+	heatAHU=${get_true_false(bldg.central_ahu.heating)},
+    coolAHU=${get_true_false(bldg.central_ahu.cooling)},
+    dehuAHU=${get_true_false(bldg.central_ahu.dehumidification)},
+    huAHU=${get_true_false(bldg.central_ahu.humidification)},
+    BPFDehuAHU=${bldg.central_ahu.by_pass_dehumidification},
+    HRS=${get_true_false(bldg.central_ahu.heat_recovery)},
+    sampleRateAHU=1800,
```
PRemmen: I think we should include parameters to the CentralAHU class instead of adding random default values to template
`sampleRateAHU`
`effFanAHU_sup`
`effFanAHU_eta`
`dpAHU_sup`
`dpAHU_eta`

teaser/data/output/modelicatemplate/AixLib/AixLib_Multizone
```diff
@@ -0,0 +1,377 @@
...
+          {36.8,-58},{36.8,-9}}, color={0,0,127}));
+
+  connect(const.y, multizone.TSetCool) annotation (Line(points={{55.2,-82},{
+          34.6,-82},{34.6,-9}}, color={0,0,127}));
+
+  annotation (experiment(
+      StopTime=${modelica_info.runtime_simulation},
+      Interval=${modelica_info.interval_output},
+      __Dymola_Algorithm="${modelica_info.current_solver}"),
+      __Dymola_experimentSetupOutput(events=${get_true_false(modelica_info
```
PRemmen: This is not the flag wo events. Add a seccond Flag to ModelicaInfo class or use this one right

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZone
```diff
@@ -0,0 +1,95 @@
+within ${mod_prj}.${bldg.name};
+model ${bldg.name}
+  "This is the simulation model of ${bldg.name}"
+
+% if zone == "ThermalZone":
+  AixLib.ThermalZones.ReducedOrder.ThermalZone.ThermalZone thermalZone(
+    redeclare package Medium = Modelica.Media.Air.SimpleAir,
+    zoneParam=
+    ${mod_prj}.${bldg.name}.${bldg.name}_DataBase.${bldg.name}_${zone.name
+	  .replace(" ", "")}(),
```
PRemmen: Check if this is already handled in thermalZone.name setter

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZone
```diff
@@ -0,0 +1,95 @@
...
+  3600
+  % for tz in bldg.thermal_zones:
+    ,${tz.infiltration_rate}
+  %endfor
+  ])
+    annotation (Placement(transformation(extent={{-92,-40},{-72,-20}})));
+
+  Modelica.Blocks.Sources.CombiTimeTable tableInternalGains(
+    extrapolation=Modelica.Blocks.Types.Extrapolation.Periodic,
+    table=[0,0,0,0; 3600,0,0,0; 7200,0,0,0; 10800,0,0,0; 14400,0,0,0; 18000,0,0,
```
PRemmen: Change this to the boundary conditions from teaser?

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZone
```diff
@@ -0,0 +1,95 @@
...
+  connect(tableInfiltrationRate.y, thermalZone.ventRate) annotation (Line(points={{-71,-30},{-40,
+          -30},{-8,-30},{-7,-30},{-7,-20},{-7,-8.4}}, color={0,0,127}));
+  connect(internalGains.y, thermalZone.intGains)
+    annotation (Line(points={{0.7,-52},{8,-52},{8,-8.4}}, color={0,0,127}));
+
+  annotation (Icon(coordinateSystem(preserveAspectRatio=false)), Diagram(
+        coordinateSystem(preserveAspectRatio=false)),experiment(StopTime=${mod_prj.modelica_info.runtime_simulation},
+      Interval=${mod_prj.modelica_info.interval_output},
+      __Dymola_Algorithm="${modelica_info.current_solver}"),
+      __Dymola_experimentSetupOutput(events=${get_true_false(modelica_info
```
PRemmen: see above

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZoneRecord
```diff
@@ -0,0 +1,281 @@
...
+% else:
+    ARoof = ${zone.area_rt},
+% endif
+    alphaRoof = ${zone.alpha_conv_inner_rt},
+    nRoof = 1,
+    RRoof = {${def_val(zone.r1_rt)}},
+    RRoofRem = ${def_val(zone.r_rest_rt-R_zero(zone.alpha_comb_outer_rt,zone.area_rt))},
+    CRoof = {${def_val(zone.c1_rt)}},
+    nOrientationsRoof = ${def_nOrientations(len(zone.tilt_rt))},
+    tiltRoof = {${list_to_string(deg_to_rad_list(def_val_Roof_list(zone.tilt_rt,def_nOrientations(len(zone.tilt_rt)))))}},
```
PRemmen: these kind of things used to be implemented in logic.simulation.aixlib/annex is there a reason of moving it to here?

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZoneRecord
```diff
@@ -0,0 +1,281 @@
...
+    CInt = {${def_val(zone.c1_iw)}},
+% if number_of_elements < 3:
+    AFloor = 0,
+% else:
+    AFloor = ${zone.area_gf},
+% endif
+    alphaFloor = ${zone.alpha_conv_inner_gf},
+    nFloor = 1,
+    RFloor = {${def_val(zone.r1_gf)}},
+    RFloorRem =  ${def_val(zone.r_rest_gf-R_zero(zone.alpha_comb_outer_gf,zone.area_gf))},
```
PRemmen: I think the calculation of resistances should be implemented into thermal zone, instead of in the templates! (also for unit test reasons!)

AixLib + Annex various occurences

teaser/project.py
```diff
@@ -80,9 +78,9 @@ def __init__(self, load_data=True):
         self.modelica_info = ModelicaInfo()
 
         self.modelica_project = self.name
-        self.weather_file_header = "wetter"
+        # self.weather_file_header = "wetter"
         self.weather_file_path = "modelica://AixLib/Resources/WeatherData/" \
-                                 "TRY2010_12_Jahr_Modelica-Library.txt"
+                                 "USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.mos"
```
PRemmen: Is this the weather file we want to use as default? given that all other TEASER modules rely on german data?

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZone
mlauster: Currently the same problem as for Annex60 models. We create the boundary condition tables in `logic.simulation.aixlib` for each building containing all zones. For single zones, we would need to adapt this to write a file per zone. For this, we would need to give this function the information what model is used. Thus, we would need to shift this function to `aixlib_output`, resp. to a output module that serves for both worlds, Annex60 and AixLib

mlauster: Should we do this in a separate issue?

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZoneRecord
mlauster: This task is shifted to #347

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZoneRecord
mlauster: What specifically do you mean? `tiltRoof`? I take the values from `logic.simulation.aixlib/annex` and  check the length, convert it to radians and to a modelica list.

teaser/project.py
mlauster: Currently we have no German weather file in the new format available in AixLib or Annex60. As soon as we integrate it, we could switch back to a German file.

teaser/project.py
```diff
@@ -1079,6 +1079,15 @@ def name(self):
 
     @name.setter
     def name(self, value):
-
-        self._name = value
-        self.modelica_project = value
+        if isinstance(value, str):
+            regex = re.compile('[^a-zA-z0-9]')
+            self._name = regex.sub('', value)
```
PRemmen: On this place we need to change also the value of `self.modelica_project` 

Please also check why there is such an attribute at all, do we really need this?

mlauster: @PRemmen, it seems as `self.modelica_project` is a duplicate of `self.name`. Would deleting this attribute require adapting the xsd schema and the bindings?

mlauster: I fixed the bug in the setter, by the way.

mlauster: Is shifted to #347

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZoneRecord
```diff
@@ -58,9 +58,8 @@ record ${bldg.name}_${zone.name} "${bldg.name}_${zone.name}"
     aExt = ${zone.solar_absorp_ow},
     TSoil = ${zone.t_ground},
     alphaWallOut = ${zone.alpha_conv_outer_ow},
-    alphaRadWall = ${zone.alpha_rad_outer_ow},
+    alphaRadWall = ${(zone.alpha_rad_outer_ow*zone.area_ow+zone.alpha_rad_outer_win*zone.area_win)/(zone.area_ow+zone.area_win)},
```
PRemmen: this might be interesting for #347

teaser/project.py
mlauster: Solved

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZone
PRemmen: #352

teaser/data/output/modelicatemplate/AixLib/AixLib_ThermalZoneRecord
PRemmen: all this stuff is moved to #347 ( #349 )

Issue #356, opened 2016-12-23
Add Continuous Integration


MichaMans:
With some major pytest improvements we should definitely integrate TravisCI or something similar.

mlauster:
merged via #357

PR #357, opened 2016-12-23
#356 added travis file to test agains 2.7, 3.4, 3.5

MichaMans:
Better late then never :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: python
+
+python:
+  - 2.7
+  - 3.4
+  - 3.5
+
+install:
+  - pip install --upgrade pip setuptools wheel
+  - pip install --only-binary=numpy,scipy numpy scipy
+  - pip install pyxb
+  - pip install -e .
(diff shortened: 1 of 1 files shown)


mlauster:
Ouh, I thought this will be one of our Nerd topics for the next year! :) What is Travis CI testing? All our test.py tests?
No objections form my side, let's merge this!

mlauster:
But it checks "only", if the tests run without error, correct? Is there a nice way to integrate tests that compare old results to new ones and state if there are any changes in the outputs at all? In this way, you can track any impact on the outputs. One way would be to have tests that do exactly that and throw errors if changes occur. Anyway, it's just an idea for the future, I will merge that! :)

PRemmen:
@mlauster You can do this by simply adding a unit test to test_data.py basically this is all what we are doing comparing "old" results to the results that TEASER outputs in its current version and if a test fails it shows the new result. Basically you can integrate what ever you want in these tests

PR #382, opened 2017-01-13
Release05

PRemmen:
This merges a lot of release features into development for testing purposes


MichaMans:
Since we merge into development and we are planning to test the next weeks, i'll merge this.
Tests are running too, should be good.

PR #384, opened 2017-01-16
Issue352 bound annex

PRemmen:


PRemmen:
I guess the decrease in coveralls is because of deletion of the assert for length of lighting in logic.buildingobjects.calculation.annex60 . As the simplified example outputs in Annex Lib do not support lighting as internal gains we do not need to check length

MichaMans:
I think this should be fine, even with this decrease.

Issue #297, opened 2016-08-17
Integration of a VDI based simulation core


MichaMans:
Thanks to @ThomasSchuetz , we now have a validated VDI 6007 based python calculation/simulation core. It would be shameful if we do not itegrate this into TEASER. 
I would suggest to integreate it in the already existing simulation module in the logic. We should take care that the calculation/simulation core is still available outside of TEASER,
Since this is a bigger upcoming integreation of TEASER i'll suggest this in MS v0.5.0 (future)

JSchiefelbein:
As requested by @MichaMans , I want to make some advertisement to move this issue to a better position on the priority list ;-) As @MiLu29 also requires a Python-implementation for thermal space heating profile generation, the VDI model integration would be a great feature.

JSchiefelbein:
We got some further support by @simonwilbertz

MichaMans:
Thanks for the support! I would suggest to do the following:
- [ ] Move weather data from simulation package to input package
- [ ] remove unnecessary files in the VDI6007 package

Maybe we could:
- [ ] Merge sun and weather files
- [ ] merge the two equal air temperatures
- [ ] integrate a function like vdi_6007_simulation() in building and thermal zone

This is just a first guess of implementing it. I'll support you as soon as possible.

JSchiefelbein:
•	[x ] Move weather data from simulation package to input package
•	[x ] Merge sun and weather files

MichaMans:
Additional Points from other issues:

- [ ] Extract houseData Dict from buildings and thermal zones
- [ ] Restructure Weather Class
- [ ] PEP8

JSchiefelbein:
Just figured out, that equal_air_temperature.py is used within validation cases, while eqAirTemp.py has been used for VDI simulation. Thus, we still need to clarify which equivalent air temperature function should be used.

JSchiefelbein:
@mlauster Still interested in participating at TEASER hack day? :-) 

4 VDI 6007 validation test cases deliver results beyond allowed temperature and/or power limitations. Seems like you are the expert for test case tuning, correct?

Failing tests:
- Case 6: assert 1.1377273146512152 <= 1 (power)
- Case 9: assert 0.11262100890958493 <= 0.1 (temperature)
- Case 10: assert 0.12689809696810173 <= 0.1 (temperature)
- Case 11: assert 5.2968144255231522 <= 1 (power)

JSchiefelbein:
[...]

PR #397, opened 2017-02-07
Issue297 vdi core pytest

JSchiefelbein:
Merge VDI 6007 validation test cases into VDI core branch #297


MichaMans:
@JSchiefelbein thanks for implementing all 12 Cases. Could you implement and run them in the test_data.py test? For now we could leave the cases not validated out of the test. but with that we should have a nice coverage for our vdi implementation from the beginning

JSchiefelbein:
I am unsure, if understood you correct: All test cases are called from the pytest environment in test_data.py. However, I left them within the verification folder. Thus, users have the option to plot validation results (and use the pytest framework)

MichaMans:
Hi, yes it would be nice if we run the py.test that the 12 validation cases (except the 4 not beeing within the validation borders) will be calculated and compared. So that we have 62 tests in our pytest called. I can't find the calls for the validation cases within the pytest_data.py

JSchiefelbein:
pytest

JSchiefelbein:
Like this?

JSchiefelbein:
Seems to work fine on issue297_vdi_core_pytest

MichaMans:
Can't reproduce this. We got 12 files changed, no work in pytest_data.py, how can i run this with py.test? Travis is also testing agains 50 cases. Thats weird 😄

JSchiefelbein:
You are right...I cannot find the current version on github, too :-D

MichaMans:
Haha ok. Then we have a hint now, it's no pushed already 😃

JSchiefelbein:
Prob. need a coffee :-P

MichaMans:
I added matplotlib to the requirements, hope travis will work now, commented the not running cases for a later implementation, maybe @mlauster could help with the validation on case 6, 9, 10, 11

MichaMans:
Mhh, Travis is still failing for Python 2.7. We need to figure out why later. For now this is fine. So keep in mind:
vdi core working with Python >= 3.4 (except cases 6, 9, 10, 11)

I'll merge it tomorrow

mlauster:
Give me some days to check the failing test cases, unfortunately 10 and 11 are tough cookies. But go ahead with merging, I will fix this afterwards.

mlauster:
Hmm, weird that test cases 3,4 and 12 fail in 2.7 . Maybe something changed in the libs used to solve the models....

PR #399, opened 2017-02-08
Issue297 vdi core test ref

JSchiefelbein:
Set new reference values to VDI 6007 test cases

### DIFF
--- a/teaser/examples/verification/vdi6007_testcases/vdi6007_case11.py
+++ b/teaser/examples/verification/vdi6007_testcases/vdi6007_case11.py
@@ -30,7 +30,7 @@ def run_case11(plot_res=False):
     """
 
     # Definition of time horizon
-    times_per_hour = 3600
+    times_per_hour = 60
     timesteps = 24 * 60 * times_per_hour  # 60 days
     timesteps_day = int(24 * times_per_hour)
 
--- a/teaser/examples/verification/vdi6007_testcases/vdi6007_case11.py
+++ b/teaser/examples/verification/vdi6007_testcases/vdi6007_case11.py
@@ -30,7 +30,7 @@ def run_case11(plot_res=False):
     """
 
     # Definition of time horizon
-    times_per_hour = 60
+    times_per_hour = 3600
     timesteps = 24 * 60 * times_per_hour  # 60 days
     timesteps_day = int(24 * times_per_hour)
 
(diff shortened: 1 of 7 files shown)


mlauster:
What is the current status? All test cases except TC11 work on Python 3.4 and additionally TC3,4,12 fail on Python 2.7?

JSchiefelbein:
Correct (according to Travis CI).

JSchiefelbein:
@ThomasSchuetz just tested pytests within Python 2.7 on his local machine. They do work (except of test 11, comparable to Python 3.5)

MichaMans:
@JSchiefelbein I'm not sure but it seems that the high resolution leads to such a high computation time that Travis is exiting the test before  the calculation ends. Does it pass the tests on your machine locally?

JSchiefelbein:
VDI 2078 limitations are only kept, if timestep is changed to 1 second (test 11). However, we can use VDI 6020 values, instead, and change the time resolution back to 60 seconds.

mlauster:
Just to see the sensitivity, you could set the tie step to 0.1 s or even smaller and see, if the results get better, maybe even hold VDI 6007. Regarding VDI 6020 thresholds, please check if TC11 is part of VDI 6020 and what the exact conditions are, e.g. checking hourly mean values?

JSchiefelbein:
@mlauster @MichaMans Okay, we should find a compromise, here. 
Seems like we are taking about two requirements:
- Enablings successful TRAVIS CI run
- Decrease time discretization to increase results accuracy

A time discretization of 1 second (or below) is exceeding TRAVIS CI runtime limit. Thus TRAVIS CI is going to fail. 60 second time discretization is exceeding VDI 6007 reference values. 

Thus, using alternative reference values (VDI 6020) and larger time discretization seems to be an alternative. What do you think?

mlauster:
For the tests it might be fine to have a larger time step, but for the justification of the approach with a larger time step and loose thresholds, it would be nice to see once if we can hold the strict VDI 6007 criteria with a smaller time step.

MichaMans:
I think we should test with Travis whats possible, so if the we run out of time with the smaller time step, we should keep this out of the travis test. In my opinion this is a powerful option in general to test such large changes in the logic (our case the vdi simulation) and it doesn't have to  do our validation (this could be done locally with a description in the testcases or documentation in general)

MichaMans:
I resolved the python 2 issues. Test Case 11 still runs into runtime timeout, but as mentioned before this must not be in the pytests, its totally fine if this is a stand alone validation case.

PR #408, opened 2017-03-14
Issue407 libraryattr

PRemmen:


PRemmen:
After rebuild all CI tests pass (ignore decrease in coverage) so I'll merge this

PR #445, opened 2017-04-18
Issue444 no int ext

### LINKED ISSUE #444: Element calculations fail if no interior walls or exterior walls but fllors or groundfloors etc. are defined
## What is the problem?

- If you define no interior or exterior walls but their horizontal counterparts, the new checks within one_element.py, two_element.py, three_element.py and four_element.py fail to catch that, what leads to errors

- To make TEASER able to calculate buildings without interior or exterior walls

- Add floor, ceiling, ground floor and rooftop to checks

mlauster:
This fixes #444


teaser/logic/buildingobjects/calculation/two_element.py
```diff
@@ -402,8 +402,9 @@ def calc_attributes(self):
                           ", please be careful with results. In addition " +
                           "this might lead to RunTimeErrors")
         self._sum_outer_wall_elements()
-        if len(self.thermal_zone.inner_walls) < 1:
-            warnings.warn('For thermal zone ' + self.thermal_zone.name,
+        if len(self.thermal_zone.inner_walls + self.thermal_zone.floors +
```
PRemmen: This applies for all three changes:

I think your changes will work just fine. 

However if I have a look at the whole function it seems that `outer walls` and `windows` are treated different. In particular following code is missing for internal walls

```python
        for out_wall in outer_walls:
            out_wall.calc_equivalent_res()
            out_wall.calc_ua_value()
        for win in self.thermal_zone.windows:
            win.calc_equivalent_res()
            win.calc_ua_value()
``` 

In addition, the function `_sum_outer_wall_elements()` is called no matter if there are outer walls avaialbe or not. This will lead to errors if no outer wall is present (which shouldn't be the case any way, but still we should handle this, right?)

```python
        if len(outer_walls) < 1:
            warnings.warn("No walls are defined as outer walls for thermal " +
                          "zone " + self.thermal_zone.name + " in building " +
                          self.thermal_zone.parent.name +
                          ", please be careful with results. In addition " +
                          "this might lead to RunTimeErrors")
        self._sum_outer_wall_elements()
```

I propose to include these changes in this PR, as they are related

mlauster: I fixed both points. I didn't know why internal walls have been treated differently, it seemed to work. Anyway, they are now treated similar and it still works. :-)
For the second point, this forced a lot of additional checks that are all added now.

PRemmen:
This branch is ready to merge

Issue #448, opened 2017-04-25
pep8


thorade:
Python has a comprehensive style guide, pep8:
https://www.python.org/dev/peps/pep-0008/
and a couple of tools to check the style.

Run the following command in the Teaser/teaser directory to get a statistic:
`flake8 --count --exclude="*/examplefiles/*" --max-line-length=120 --statistics -qq`

Then, to find the files and lines with style violations run something like
`flake8 --count --exclude="*/examplefiles/*" --max-line-length=120 --select="W291"`

And then, to automatically fix (if possible) run something like
`autopep8 --in-place --recursive --max-line-length=120 --exclude="*/examplefiles/*" --select="E101,W1,W2" .`

The fixes that only change whitespace are probably safe to apply, and should improve readability of the code.

A line length of 120 should be the absolute maximum:
http://stackoverflow.com/q/22207920/874701
shorter lines of 100 characters are preferred e.g. by pylint,
pep8 even defaults to 79 characters!

PRemmen:
@thorade the very most of pep8 violations should be resolved in the pull request #449. You are invited to review. The few remaining issues are (as far as I see) related to citygml pyxb binding classes, we need to import all pyxb modules although we do not explicity use them (this is a constraint of pyxb) and a small workaround for export in `logic.buildingobjects.building` which will stay like this.

If you're running statistics please remember to exclude the folder `teaser/data/bindings` as already mentioned in the PR #449  these are autogenerated files where we don't want to apply `autopep8` (in addition, I do not see any reason why these files should be human readable)

Hope your reading experience of TEASER will be improved with this PR ;-)

thorade:
> Hope your reading experience of TEASER will be improved with this PR ;-)

A lot, all that surplus / missing whitespace was driving me crazy!

thorade:
Just another comment: If you happen to use Spyder as your editor, you can analyse the code quality very conveniently by just pressing F8.

![spyder_codeanalysis](https://user-images.githubusercontent.com/314571/27372368-ad3e41ca-5664-11e7-852a-71f825f8b4bc.png)

PR #449, opened 2017-04-25
Issue448 pep8

PRemmen:
Applied autopep8 for `W391, W293, W292, W281` as suggested in #448 . Autogenerated bindings do have a lot of errors, although there are all white space errors and should be okay I'm not going to apply auto pep8 on autogenerated files 

solves #448


thorade:
flake8 returns non-zero when it finds something, so it can be used in the travis.yml to test compliance to pep8. Only include few rules that are wanted, and exclude the autogenerated files.

PRemmen:
:+1: thanks @MichaMans I'll merge

PR #462, opened 2017-06-21
Issue439 tabula project

PRemmen:
This merges TABULA archetype buildings for germany into development. As this implies some main code changes, espacially for `data` package. I suggest we merge it into development and test it for some weeks until we make a new release and push it to master.

I will also create a table with all constructions used. as this is not code related we can do a separate pull request.


PRemmen:
@MichaMans I added tests for backwards compatibility of teaserXML, now coverage should be ok

PRemmen:
@MichaMans I created wiki page for the wall constructions, so from my point of view ready to merge

Issue #466, opened 2017-07-11
Add examples for TABULA archetypes


PRemmen:
## What is the problem?

- I forgot to update the example that explain how to use tabula archetypes

## Why do we want to solve it?

- to explain new users the usage of TABULA archetypes

## How do we want to solve it?

- adding examples :-)

PR #469, opened 2017-07-12
Issue466 tabula example

PRemmen:
This solves #466 and adds examples for TABULA. In addition it solves problems with retrofitting tabula buildings and extends the retrofit tests


PRemmen:
I will add some tests to cover missing lines in project.py

MichaMans:
@PRemmen should we add this test for now? And maybe implement further tests for increasing coverage? At the moment this should be fine for me

MichaMans:
a slightly decreasing coverage will be fine. Should we merge this? @PRemmen

PRemmen:
thanks for approving @MichaMans

PR #470, opened 2017-07-14
Update setup.py

thorade:
as mentioned here, but automatic
https://github.com/RWTH-EBC/TEASER/releases/tag/v0.5.0
also see
https://stackoverflow.com/a/8161816/874701
https://docs.python.org/3.5/distutils/setupscript.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,4 +97,4 @@ setup(name='teaser',
           'Topic :: Software Development :: Code Generators',
           'Topic :: Scientific/Engineering',
           'Topic :: Utilities'],
-      install_requires=['mako', 'pyxb>=1.2.5', 'pytest', 'scipy'])
+      install_requires=['mako', 'pyxb==1.2.5', 'pytest', 'scipy'])


setup.py
```diff
@@ -97,4 +97,4 @@
           'Topic :: Software Development :: Code Generators',
           'Topic :: Scientific/Engineering',
           'Topic :: Utilities'],
-      install_requires=['mako', 'pyxb', 'pytest', 'scipy'])
\ No newline at end of file
+      install_requires=['mako', 'pyxb>=1.2.5', 'pytest', 'scipy'])
```
PRemmen: Hi @thorade ,

thanks for that pull request. However, I won't accept it ;-).

pyxb provides no backwards compatibility. That means, a binding (e.g. CityGML or TypeElements in TEASER) generated with pyxb version 1.2.5 won't work with pyxb version 1.2.6. If there is a new release of pyxb ,a new installation of TEASER wouldn't work without updating the bindings. To conclude: we actually always need a distinct verison of pyxb, So changing your code to `pyxb==1.2.5` would add value.

Issue #117, opened 2016-01-22
Adjacent rooms


mlauster:
Allow the definition of multiple exterior walls that are connected to multiple adjacent rooms with defined temperature, preferably as profile. This would force changes of `EqAirTemp` Model in AixLib as well to allow adjacent rooms (one vector with weightfactors and one with temperatures or connector to profiles).

InaDJ:
@mlauster
Hi Moritz, 
I have been using TEASER to translate CityGML files into IDEAS models (I told you I was planning to do that when we were in Corsica). Now, I do have a problem and I would like to hear your view on this. Within the CityGML files, all buildings are represented by all of their wall surfaces (shared walls are thus two overlapping surfaces). Treating these shared walls as external walls means a significant overestimation of the energy use of a terraced house or even a semi-detached house. I am aware of your solution for residential buildings, but this does not apply when importing a CityGML file, right? Therefore, I implemented in TEASER the possibility to "detect" shared walls and delete them. Is this something you plan to include in TEASER? 
My approach seemed to work well, until I tried it to apply it to a more detailed CityGML model and then it didn't detect all shared walls anymore. I'll explain my approach in short: if you import a CityGML file in TEASER, TEASER takes every surface and creates either a Wall, Roof or Groundfloor for it. After these are all created, I loop all these surfaces again, looking for an overlapping surface (I did this by determining the "equation" of the surface, based on the normal vector and one point). If there is an overlapping surface, I calculate their common area (using another Python package (Shapely, I think)) and subtract it from the original area. This is equal to just deleting the overlapping part (by which you also lose its thermal mass). In case the resulting area is zero, I just delete the Wall-object. But as I said, this approach does not work yet in all cases (I think that my surfaces are really overlapping in CityGML, so probably it is due to rounding errors while determining the equation of the surface (issue of the floating point error!)). 
I think detecting these shared walls would be an enhancement for TEASER, but the user should be able to choose if he'd like to detect these or not, as it could increase the calculation time (scanning 100 buildings for adjacant surfaces took about a minute or two tops in my case, if I remember correctly). Instead of deleting them, we could "label" them as internal walls or create a new type of walls?
In case you are not planning to include this in TEASER, I won't look for a solution to my problem any further and try to just delete the shared walls before creating my CityGML models and then importing these improved ones into TEASER. 
Looking forward to hear your opinion on this, 
Kind regards, 
Ina

mlauster:
[...]

PR #473, opened 2017-07-19
Issue117_adjacentbuildingscitygml

InaDJ:
#117 Adjacent rooms
This adds the possibility to delete shared surfaces with adjacent buildings when importing a cityGML file. For now, only exact matching surfaces are addressed (no tolerance on surfaces). I added some print statements to the building.py script in order to make more clear what I did, these should of course be deleted. (Within the code, there are also quite a lot remarks which should it more readable, feel free to delete these). 
What I did: 
After the project and all buildings (and buildingelements) are created, this "reset_outer_wall_area" function (in building.py) is called from the load_gml function for every vertical gml_surface. This functions checks if any other gml surface of the project is coplanar with the considered one. If it is, the common area is calculated. If this common area differs from zero, then the window object in this direction is deleted (assumption: no windows in shared walls!), its area is added to the corresponding wall, then the common area is substracted from this wall. If the remaining wall area is very small or nan, then the wall object is deleted. 
I'm sorry, there is still an error (zerodivisionwarning), I did not solve this one, since it is not important for the work I do :)
Please, feel free to share any thoughts you have on this ! I'm happy to contribute!

Kind regards, 
Ina


mlauster:
@InaDJ, sorry for the delay while reviewing your code, there were some pressing issues to be solved beforehand. However, I had a first look at your code. Thanks a lot again for your work and and participating on TEASER. I still think your module could be a valuable extension!
To start with the review, travis is having some issues with your code regarding PEP8 style. You can view the details when clicking on "Details" next to the travis annotation. Just choose one of the jobs and have a look at the job log. You will find 43 PEP8 issues.
I saw that you already started working on an implementation without exact matching. Above you wrote, so far only the exact matching works. What is your current status on this? I think it would be very valuable to have the non-exact matching as I assume that this is a common case.
In addition, it would be great to have a small example that contains buildings with different adjacency, e.g. exact matching, non-exact matching regarding areas, non-exact-matching regarding angles. That would make it easy to assess your code and see if it works as expected. 
Afterwards, we could discuss the details of your code and get it `development`-ready!
By the way, in my opinion, with a bit of code polishing and extension of functions, this is a nice topic for a paper of you!

InaDJ:
@mlauster, I'm sorry as well for the delay. I was on holidays and when I came back, I had multiple deadlines (with the last one on the 1st of November), so I'll further look into it in November. I did not complete the implementation without exact matching as it did not seem to work out and I decided to "pre-process" my CityGML files some more in FME (the software that I use to create them in the first place based on ESRI shapefiles). Since I make sure that the "not exact matching surfaces" are exact matches in FME, I decided not to finish the code. I only included my first attempt to serve as a first idea, in case you would like to actually implement it. However, I don't think that I will further elaborate on this, since the current version works fine for my goals and I kind of have other priorities.  Additionally, since this summer, I have been processing larger neighbourhoods and as expected, the time to "look for adjacant surfaces" increases drastically with the size of the neighbourhood. We could solve this by adding "an additional filter", before checking that the surfaces are coplanar, so first check if they are close to each other.  Conclusion in short: I am not sure if the current version is ready to be implemented in the development branch, as there are still some large drawbacks, but I do think that such a feature would be nice for TEASER (even if it is not the main focus).

Kind regards, 
Ina

mlauster:
@InaDJ, thanks for that update! And again thanks a lot for your input, regardless if you can afford to improve your approach or not. Under the impression that this part of code is not yet production-ready but contains a lot of potential to be a valuable extension of TEASER, I would close that PR and keep issue #117 open as well as keep the corresponding branch. If I have the chance to assign a student, I will pick up that issue again. However, if you come across that part of your work again and fix that "non-symmetrical" stuff, please let us know and we could reopen the PR!

InaDJ:
@mlauster , thanks a lot for your kind answer. I totally agree with the fact that the code is not production-ready. To have a student working on it would indeed be a great idea. If I ever work on an update, I'll let you know. 
Kind regards, 
Ina

PR #477, opened 2017-09-04
increase version number

PRemmen:
This merges the increase of the version number.


MichaMans:
Thanks Peter! Could you create a tag with that? :+1:

PR #481, opened 2017-09-08
Issue297 vdi core jsc 2

MichaMans:
With this PR we have a new working copy of the vdi implementation. Please delete unnecessary branches! 
I'll delete issue297_vdi_core_weather @JSchiefelbein could you please have a look at the _jsc branches and delete the not in use ones?


MichaMans:
Note: Checks will fail cause of the deleted packages. We need to use the new core for the tests, i'll start working in them after mergeing

JSchiefelbein:
Deletet old development branches. 
However, issue297_vdi_core_dev_jsc should remain, for the moment, as it is the only branch, which currently enables usage of complex city pickle files for optimization in pyCity_opt

Issue #136, opened 2016-02-16
Adjust required AixLib version


MichaMans:
This can be reopend every time the AixLib version changes. If so, we need to adjust the supported AixLib version in Teaser

mlauster:
Closed via #489

MichaMans:
As the AixLib released in v0.7.1 we should adjust that in teaser

PR #489, opened 2017-10-12
Issue136 aix lib version

mlauster:
This is to update the version number of AixLib within TEASER to V0.5.2, what is the current version of AixLib in `master`. Since this is some kind of bug and annoying for the user, I would merge it into TEASER `master` and increase the version to V0.5.4. This will then of course force a new release, what I will do after the merge.


mlauster:
Ah, and we need to close #136 again afterwards.

mlauster:
IBPSA is still up-to-date. Where do we need to adjust the AixLib in calculation.ibpsa? That would make no sense at all...

MichaMans:
I thought we might adjust it here:
teaser/logic/buildingobjects/calculation/ibpsa in the init? Does the IBPSA TEASER modelsupport not need the parentlibrary version?

mlauster:
No, that's fine, it only contains the version number of the ibpsa library, which is still up-to-date.

MichaMans:
Ahh sorry, i figured out what my problem was / is. I was looking at the development branch. There we specifiy the Parent Library versions in the mentioned ibpsa class. So this here is all fine. #492 might be an enhancement nevertheless.

mlauster:
Ah, I saw an instant ago as well! Then we should fix this at frist, merge it into `master`, merge `master` into `development`and then check all other libraries.

MichaMans:
Yes. We should fix it in the master by merging this pr. then we see how we will handle the future version in the development

Issue #491, opened 2017-10-12
Create test for the actual release versions of the different supported modelica libraries


MichaMans:
## What is the problem?

Supported modelica libraries are changing over time

## Why do we want to solve it?

We should support always the latest version of the libraries

## How do we want to solve it?

create a unittest which trys to import PyGithub and if it finds it, it should test for the supported versions

mlauster:
Ouh, could there be an automated way? Using the release numbers or how does it work? This would really be a huge advantage!

MichaMans:
solved with PR #492

PR #492, opened 2017-10-12
#491 added a test for the library versions

MichaMans:
after #489  is merged and the library versions are ajusted we can merge this and we get an automated test s for the library versions in the future.

Feedback welcome @mlauster @PRemmen


MichaMans:
Ok, travis failing, we need to add a github token for the request. Might be a good solution. I'll create one tomorrow

MichaMans:
Ok. Seems to work now. We need to adjust the libraries then and test again and see if travis fails. But all after #489 is merged

MichaMans:
I test again the tags, releases on Github. Not the modelica code. Thats a good point! I can think about pulling every repo and test the version in the code. We could discuss this in #491 . For a first version this should work.
What i do:
- Install PyGithub (Python Github API)
- Get the latest releases of the modelica libraries (With a secret github token via travis_ci environmental variable)
- Compare them to the code in TEASER

mlauster:
@MichaMans, what about merging this? :)

MichaMans:
@mlauster i updated to the latest buildings library. Could you have a quick look and approve this?

mlauster:
@MichaMans, go ahead and merge!

Issue #506, opened 2018-04-24
Bug: typeBuildingElements only defined until year 2015


anna-scholl:
## What is the problem?

-  in `TypeBuildingElements.xml` the elements are only defined until year_of_construction 2015 

## Why do we want to solve it?

- For year of construction >2015 no TEASER buildings can be created, because building_elemnts are None

## How do we want to solve it?

- Maybe the year in  `TypeBuildingElements.xml` can be changed from 1995-2015 to 1995-2020 or something like this.

anna-scholl:
It would be nicer to change the year of construction to year 2024, since in my project I have buidinigs which are planed for 2024

PRemmen:
@anna-scholl thanks for this issue! this is indeed not very nice :). Do you need help solving this?

PR #516, opened 2018-06-01
Changed typeBuildingElement date to 2030 to create teaser buildings for future year_of_construction 

anna-scholl:
Only typeBuildingElement year is changed to 2030.
This solves #506


PRemmen:
@anna-scholl thanks for the PR. I'm not sure if this is the best solution. As users may think the buidling standard holds true for the whole period, which is kind of misleading. I'd rather keep the current form but integrate a user warning if the input is higher than 2015 that says that the building standard from 2015 will be used.

Off topic: we need to make decision on how to handle those old building elements. As we know have the ones from TABULA which should be in general more accurate.

Issue #523, opened 2018-06-28
Bug in generate archetype


MichaMans:
## What is the problem?

atm the generate archetype function in the singlefamily archetype has a critical bug. The number of persons loaded from the zone usage data is not multiplied with the zone area and divided by a "normperson". 

## Why do we want to solve it?

The current implementation is leading to a not changing number of persons in the residential buildings

## How do we want to solve it?

use the same method like in the office or generate_from_gml function

MichaMans:
should be solved in masterbranch, nevertheless we need to to this for the tabula archetypes in the development as well
- [ ] fix bug in tabula buildings

PR #522, opened 2018-06-28
Set self.sum_heat_load to zero (building.py / Building object) to pre…

JSchiefelbein:
Bugfix for #518. Sets sum_heat_load attribute to zero before retrofit of building.
Closes #518

PR #524, opened 2018-06-28
#523 fixed bug in residential buildings

MichaMans:
This solves #523


PRemmen:
approved!

maybe release it together with #522

PR #525, opened 2018-07-03
#523 added number of persons and machins calculation

MichaMans:
Should solve #523 completely


PRemmen:
this is okay. I'll try to fix the issues with not passing py tests today.

Issue #520, opened 2018-06-25
Buildings Release 5.1.0 breaks CI tests


mlauster:
## What is the problem?

- Buildings lib at https://github.com/lbl-srg/modelica-buildings/releases has been released in version 5.1.0. We currently test for version 5.0.1. This leads Travis CI to break tests

## Why do we want to solve it?

- To be able to test and develop TEASER further again

## How do we want to solve it?

- Add new version number

PRemmen:
only in development currently, correct?

PRemmen:
strange enough, this only fails for py 2.7 I'll have a closer look another time

mlauster:
Ah, got it, you meant your fix.

PRemmen:
@MichaMans I have now idea why this fails on python 2.7, suspect it has something to do with PyGithub, can you help me here?

mlauster:
The test complains about encoding this character: http://www.fileformat.info/info/unicode/char/1F423/index.htm, what is a hatching chick pictograph, and really might be a problem with Python 2.7. It only appears for IDEAS, so maybe there is something strange in one of their commits...

MichaMans:
@mlauster maybe we should test fixing this issue in the issue520 branch? Already set pygithub to version 1.39 which solves the issue. Maybe the issue occurs because they changed from httplib to requests.

mlauster:
hm, I get a new error for 3.5 and 3.6...
https://travis-ci.org/RWTH-EBC/TEASER/jobs/400043300

mlauster:
Got it fixed by restarting the job :-)

PR #526, opened 2018-07-04
Issue520 buildings test

MichaMans:
I think this solves #520 

@mlauster @PRemmen Are you fine with setting PyGithub back to a specific version as long as this one does the job for our small test?


mlauster:
Fine for me!

Issue #537, opened 2018-09-20
Bugfix for tabula windows


PRemmen:
There is a bug with g-values in the Tabula type elements, I'll try to fix it today and then we can release version 06.2. and also release new pip version

Issue #505, opened 2018-01-09
Bug: BuildingElement with layer thickness of 0.0


PRemmen:
## What is the problem?

- in `TypeBuildingElements.xml` the `OuterWall` 1995-2015 "light" has two layers with thickness of 0.0 

## Why do we want to solve it?

- it produces errors in records and thus can't be simulated

## How do we want to solve it?

- add proper thickness to the materials in XML file, before double check with older versions of TEASER to get right thickness

MichaMans:
Maybe we should interate this with #538 ?

PR #538, opened 2018-09-20
fix windows bug

PRemmen:
solves #537 , please also check if new pip is released! 

values are according to calculation sheet of tabula


PRemmen:
@MichaMans something we can merge? :)

PRemmen:
@MichaMans just read your comment on #505 I cherry-picked the corresponding commit

Issue #544, opened 2018-10-26
Integrating Advanced Building Simulation


DaJansenGit:
During the B13 Project we want to integrate some new features into TEASER. This issue should be used as a roadmap for different implentations. Further Information will be added.

## What is the problem?

- ...

## Why do we want to solve it?

- ...

## How do we want to solve it?

- ...

MichaMans:
- issue544_aircurtain introduces the air curtain parameterization

DaJansenGit:
issue544_RetailProfiles integrates weekly profiles for persons, lighting and machines for retail buildings to take weekends into account.

Issue #510, opened 2018-05-02
Clean Up UseConditions


PRemmen:
## What is the problem?

- we drag around a lot of variables we actually do not really need for use conditions

## Why do we want to solve it?

- have it more compact and user friendly

## How do we want to solve it?

- delete all unused parameters

PRemmen:
Special focus should be on internal gains, especially machines and persons as this is currently kind of messed up in TEASER

MichaMans:
Some collection for discussion and or implementation:

- persons: Currently implemented as W/m² with one person at 24 degreeCelsius and 95 W per persons. 
  *Attention:* Current implementation only says 70 per persons

DaJansenGit:
regarding the persons:
the current calculation doesn't scale with the zone area. What we want to integrate is simple:
person_density: nr of people / m² in zone

this is then scaled with the zone area and used as "NrPeople" in the HumanSensibleHeat_VDI2078 modelica model.

PRemmen:
I'll start to clean up use conditions. My plan is the following and is also related to other issue (#539  #578 #508 #438).

1. I will delete all unused parameters from UseConditions, this should improve usability of the existent ones
2. I will no longer use XML for saving UseConditions in our internal database, but rather use JSON. 
3. I will change to schedules in pandas (I will also integrate some conversion functionalities). however, this means that TEASER will no longer support heating time and the current logic to implement a night set back.

This is going to be a major change and probably will take me some time and also the plan might change will implementing. However I think this is really a good improvement

PRemmen:
I've made some first implementations, one thing that we really should consider is, that we will not be backwards compatible (*.teaserXML) any more without huge effort. this holds especially true for all the parameters that have been loaded withour any use and also because I#m now using UseConditions directly instead of BoundaryConditions (which will be deleted).

PRemmen:
Just that we do not forget about it: moved infiltration_rate from thermal_zone to use_conditions.

PRemmen:
@MichaMans the unit tests for VDI calculation do not work although they work locally. Do you have any idea why?

MichaMans:
@PRemmen could you please be a little bit more specific? Whats the vdi calculation test and what exactly is failing?

PRemmen:
@MichaMans 

https://github.com/RWTH-EBC/TEASER/blob/31adfa7f4109920729e6ea24a82d5e0e7f1f995e/tests/test_data.py#L21
[...]

PR #556, opened 2018-12-12
Allow for AHU profiles of different lengths

### LINKED ISSUE #553: AHU profile for durations other than 1 day
## What is the problem?

For an office zone, I would like to specify different volume flow profiles for workdays and for the weekend. Just passing a list of length 7 * 24 to `building.central_ahu.profile_v_flow` does not work, because when `export_aixlib()` calls `modelica_AHU_boundary()`, it runs 

```Python
        if time_line is None:
            time_line = self.create_profile()
``` 

and `create_profile()` uses its default parameters `create_profile(duration_profile=86400, time_step=3600, double=False)`. 

- In my opinion, it would be useful to allow users to specify different AHU operation schedules for periods other than 1 day repeated for the entire year.

- I am just playing around with different use_conditions, so it's likely that I am just approaching the problem from a wrong angle.
- If there is no other way already implemented to set AHU operation for workdays and weekends, I suggest to not use the default value of `duration_profile=86400` in `create_profile()` , but instead simply determine the length of  `building.central_ahu.profile_v_flow` and create a corresponding `time_line`.

## Next steps

- If someone (@PRemmen , @MichaMans ?) could provide some feedback whether this is indeed a problem or whether there is a different approach to solving my problem and a solution is needed, I could implement the suggested solution.

marcusfuchs:
Closes #553. Change the AHU profile processing to work with different lengths for the profiles, as long as the longest profile's length is a multiple of all the shorter profiles' lengths.

@DaJansenGit @nhuelsenbeck : This seems to be related to your work in #544, too, so it would be great if you could also take a look at this.

### DIFF
--- a/tests/test_data.py
+++ b/tests/test_data.py
@@ -3159,3 +3159,71 @@ class Test_teaser(object):
             assert ideas.replace('v', '') == ibpsa.version['IDEAS']
         except IndexError:
             warnings.warn('There was an index error for IDEAS', UserWarning)
+
+    def test_ahu_profiles(self):
+        """Test setting AHU profiles of different lengths
+
+        Related to issue 553 at https://github.com/RWTH-EBC/TEASER/issues/553
+        """
+
+        prj_test = Project(load_data=True)
+        prj_test.name = "TestAHUProfiles"
+
+        prj_test.add_non_residential(
+            method="bmvbs",
+            usage="office",
+            name="OfficeBuilding",
+            year_of_construction=2015,
+            number_of_floors=4,
+            height_of_floors=3.5,
+            net_leased_area=1000.0,
+        )
+
+        prj_test.used_library_calc = "AixLib"
+        prj_test.number_of_elements_calc = 2
+
+        v_flow_workday = [
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            1.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+        ]
+
+        v_flow_week = []
+        for day in range(7):
+            for val in v_flow_workday:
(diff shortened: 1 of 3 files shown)


tests/test_data.py
```diff
@@ -3159,3 +3159,71 @@ def test_modelica_export_version(self):
...
+                if day < 5:
+                    ratio = val
+                else:
+                    if val == 1:
+                        ratio = 0.2
+                    else:
+                        ratio = 0.0
+                v_flow_week.append(ratio)
+
+        for building in prj_test.buildings:
```
MichaMans: Could we test some kind of original values with the values implemented by hand to be sure that the same things are happening?

marcusfuchs:
@MichaMans Thanks for the feedback. I must admit that it has been a while since I worked on this, so I will have to re-familiarize myself with the issue. I think I also saw that somewhere else in the code there was a similar setup for profiles but handled separately and a bit differently. I'll try to have another look at the code and your comments. Yet, unfortunately I cannot make time for this immediately, so it may take a while for me to re-visit this.

PRemmen:
@marcusfuchs I'll did something very similiar for heating, cooling and internal gains profiles. In my opinion my approach is a bit more simple and also better to maintain:

The workflow is as follows:

1. pass over a list of any length (e.g. one day, one week, one month) with the desired values
2. the setter will use this list to fill a dataframe in UseConditions for a whole year: https://github.com/RWTH-EBC/TEASER/blob/issue510_usecond/teaser/logic/buildingobjects/useconditions.py#L134
3. despite some definition of dataframe the generation for the output files (now all txt) are just a few lines https://github.com/RWTH-EBC/TEASER/blob/issue510_usecond/teaser/logic/buildingobjects/calculation/aixlib.py#L184 

I propose that I'll adapt that also to the AHU profiles. From what i see this should not "destroy" anything in your workflow.

marcusfuchs:
@PRemmen Great suggestion. Back when I proposed this change, I needed a fast solution for my special application. I only later realized that there are many other parts of TEASER where a similar solution would help and that my solution does not cover all cases gracefully. Therefore I am perfectly happy with closing this PR and adapting to your solution.

PRemmen:
This issue will be handeld in issue510_usecond #510

Issue #561, opened 2019-02-06
Rotate building working correctly?


marcusfuchs:
I am trying to rotate a building with 

```python
bldg = prj.add_non_residential(
        method="bmvbs",
        usage="office",
        name=f"East",
        year_of_construction=year,
        number_of_floors=2,
        height_of_floors=3.5,
        net_leased_area=1000.0,
        window_layout=2,
    )
bldg.rotate_building(angle=90)
```

Now I am trying to verify whether the rotation is working correctly. When looking at e.g. the `Office` zone's data record, without the rotation I get the following external wall areas:

```modelica
AExt = {55.95947018972092, 287.5, 55.95947018972092, 16.447218194891885, 287.5, 16.447218194891885},
```

I was not able to find a definitive documentation for the order of orientations in this dataset, but I would expect the 2nd and 5th item (`287.5`) to define the area of the roof and ground, while the other 4 elements would be the 4 wall orientations. 

When I rotate the building by 90 °, I would expect to get something like this, where the 4 wall orientations have switched places:

```modelica
AExt = {16.447218194891885, 287.5, 16.447218194891885, 55.95947018972092, 287.5, 55.95947018972092},
```

Instead, what I get is this:

```modelica
AExt = {16.447218194891885, 287.5, 287.5, 55.95947018972092, 55.95947018972092, 16.447218194891885},
```

Here, the larger areas which I assume to be for roof and ground have different positions. My guess would be that this is not correct. But maybe some other place in the record or model is also updated and this is what should be expected? Could someone with knowledge of this function help to verify whether this is an error or expected behavior?

PRemmen:
I'd say not necessarily wrong. Depends on what you get in your records for: 

```modelica
tiltExtWalls
aziExtWalls 
```
The order of the lists are always used in combination.


```modelica
AExt = {390.73567665235294, 646.875, 390.73567665235294, 51.04102752019142, 646.875, 51.04102752019142},
aziExtWalls = {0.0, 0.0, 3.141592653589793, -1.5707963267948966, 0.0, 1.5707963267948966},
```

```modelica
AExt = {51.04102752019142, 646.875, 646.875, 390.73567665235294, 390.73567665235294, 51.04102752019142},
aziExtWalls = {0.0, 0.0, 0.0, -1.5707963267948966, 1.5707963267948966, 3.141592653589793},
```
here you see that the ares are rotated correctly.

[...]

PR #563, opened 2019-02-06
treat 360 as 0 degree

PRemmen:
should solve bug reported in #561


marcusfuchs:
Thanks for the very fast fix! I can confirm that also in my test model, this solves #561 .

PRemmen:
@marcusfuchs thanks for the review. Just one small request. It would be nice if next time you just approve the PR and then let us merge the branch. E.G. this time i forgot to increase version (thanks @MichaMans !!!) and thus our CI for PIP release is not triggered :)

marcusfuchs:
Sure, no problem. I think last time (#558) I was asked to merge directly, but I can just review next time and leave the merging to you.

PR #579, opened 2019-04-17
Switch from facade area estimation to calculation

miraschmid:
With this implementation, we are now calculating the facade area if the required information is available. Furthermore, we are now using 3 as a default value for the office_layout instead of 0.

#569


marcusfuchs:
@PRemmen @MichaMans :

@miraschmid continued the work I started for #570. I'll merge this to my branch so that we can continue the discussion in the already existing PR.

PR #586, opened 2019-05-10
Stop support of older Python versions

### LINKED ISSUE #585: Drop Python 2 support
The clock is ticking: https://pythonclock.org/

In order to make development easier, I would suggest to stop testing for and supporting Python 2 as soon as possible. I would even go further and drop support for anything < 3.6 . It's an easy simplification of the development process and I would expect this to not affect any users?

My ideal outcome of this issue would be to drop support for anything below Python 3.6 immediately. If this is not possible, could we at least have a plan on when to stop supporting Python 2?

marcusfuchs:
Closes #585. As there was broad agreement to drop support for any Python versions below 3.6, This removes the automated tests for older Python and updates the README accordingly.


MichaMans:
@marcusfuchs could you delete the tests in the .gitlab-ci.yml as well 😃 whould be great 👍

marcusfuchs:
@MichaMans Right, I was not aware of you running both CIs in parallel. The gitlab CI includes a duplication of the 3.6 test, which I'll also delete.

MichaMans:
@marcusfuchs  perfect! Take care! It is not a duplication 😃

marcusfuchs:
@MichaMans Thanks, should all be fixed now as discussed.

MichaMans:
@marcusfuchs  ok. reason for the duplication of the py3.6 test was that one is only tested on master. The one with the modelica version tests. All other tests are ignoring this because there might be issue with the different modelica versions as we agreed on supporting modelica libraries --> teaser master  and aixlib development --> teaser development.

Sorry for not sorting it out directly, but i was not thinking deep enough in it.

Issue #588, opened 2019-06-13
IdealHeaterCooler Outside temperature controlled


DaJansenGit:
With the integration of https://github.com/RWTH-EBC/AixLib/pull/732 into AixLib we have a controller that sets the ideal heater cooler active or unactive by taking the outside air temperature into account. For TEASER we need to integrate some parameters for export to parameterize the controller. 
This will be:

- _zoneParam.use_conditions.T_threshold_IdealCooler_ Threshold temperature for cooling (default should be 24 °C)
- _zoneParam.use_conditions.T_threshold_IdealHeater_ Threshold temperature for heating (default should be 15 °C) 
- _zoneParam.use_conditions.IdealHeaterCoolerThresholdControlled_ Should controller be used: if yes: controller will be used and "statorDyn" will be set false for the heatercoolerPI

PRemmen:
probably related to #617 and #618

PR #589, opened 2019-06-14
WIP: IdealHeaterCooler Controller by outside temperature

DaJansenGit:
This PR is related to #588 
The needed parameters to parameterize the controller are added to the export for AixLib ROM.


DaJansenGit:
Since @PRemmen ist changing the filesystem from the use conditions to json in #584, I will wait for the integration and integrate the default values into the use conditions.

DaJansenGit:
we don't use this in B13 but I opened a second branch and PR to bring this into TEASER/AixLib here #618 
I will close this PR.

Issue #567, opened 2019-02-26
Adjustment of internal gains calculation


nhuelsenbeck:
## What is the problem?

- The exported variables nrPeople and nrPeopleMachines need to be changed to specific variables (W/m^2).

This issue is related to https://github.com/RWTH-EBC/AixLib/issues/690

## Why do we want to solve it?

- To increase the transparency of the interal gains calculation.

## How do we want to solve it?

- Remove the 0.01 factor in every archetype

- Change name of nrPeople and nrPeopleMachines in ThermalZoneRecord file

DaJansenGit:
This should also solve the problem explained in #523 for non archetype buildings.

PRemmen:
@DaJansenGit sorry to be a little pedantic here :)

1. could you please check if the documentation and the names of paramters in TEASER classes are still correct?
2. have you compared simulation results? are they 100% the same?
3. aren't here any changes in UseConditions.xml necessary?

thanks!

DaJansenGit:
All fine, thanks for being pedantic @PRemmen  ;)

1. we just added a minor additional comment in the documentation, but now it should be fine
2. the results aren't 100% the same. But this was expected due to the former workaround that used a fixed multiply by *0.01 in TEASER (by expecting an heat transmission of 100 W/person) but a non fixed calculation in modelica, where the heat transmission of people is depending on the room temperature and activity type. Due to the fact that this activity type information and the temperature dependence is not allocated in SIA 2024 or DIN V 18599-10 we removed them. I hope you can follow what i tried to say but in the end it should mean: we have slight differences in the results but the new results are the correct ones ;)
3. Nope the UseConditions.xml don't have to be touched because we still use the same data from DIN V 18599-10 as before.

PRemmen:
@DaJansenGit @MartinRaetz thanks for clarification 👍 

can you give just one example what not 100 % the same means :) But then I'll really keep my mouth shut.

DaJansenGit:
@PRemmen 
You can find an example here https://github.com/RWTH-EBC/AixLib/pull/691 ;)

PRemmen:
do we have one example what this implies on the total energy demand and the dynamic behaviour of the thermal demand/indoor air? maybe @nhuelsenbeck has the  results files still available. Don't get me wrong: I like the adjustment a lot, however I'm running some simulation for different projects so i need some estimation of the effects :)

and then: :no_mouth:

MartinRaetz:
[...]

PR #594, opened 2019-07-11
Issue567 adjust of internal gains from dev

DaJansenGit:
This PR solves the TEASER site of the issues mentioned in #690 and in RWTH-EBC/TEASER#567


MichaMans:
@DaJansenGit Could you please have a look again and check with PR #600 that we update the branches and merge them accordingly afterwards. I would suggest to merge #600 first as this is a smaller thing. Then update this one here and i'll do the review

PRemmen:
@martinkremer @DaJansenGit Thanks a lot for this PR! 

I just had a quick look. Currently I'm seeing to difficulties:

1. You haven't introduced the new attributes into load and save function (both use_conditions and teaser project) - I can help with that, as this is also omething for #602 
2. I'm not sure of json operation like {"/":[1,14]} can be interpreted with the normal json parser in python, can anyone verify this? :)

DaJansenGit:
> @martinkremer @DaJansenGit Thanks a lot for this PR!
> 
> I just had a quick look. Currently I'm seeing two difficulties:
> 
> 1. You haven't introduced the new attributes into load and save function (both use_conditions and teaser project) - I can help with that, as this is also omething for #602
> 2. I'm not sure of json operation like {"/":[1,14]} can be interpreted with the normal json parser in python, can anyone verify this? :)

to 1.: I was not aware, that we need to change the load and save functions, I will have a look at the implementation and will contact you if I need help @PRemmen ;)
to 2.: Just checked the syntax, this is ok, but you are right! I will test if the python json module supports arithmetic operations.

Thanks for your help!

DaJansenGit:
Just checked the json export, by default python doesn't parse the division correctly, instead it gets exported as an ordered dict, you can see here:
image
With 62d3005 I integrated a workaround, is this ok for you @PRemmen

DaJansenGit:
I think most of the work is done now, I also made all necessary changes in Modelica (see https://github.com/RWTH-EBC/AixLib/issues/695). Let's see if the tests are passing, if not I will have a look again.

PRemmen:
@DaJansenGit @KremerMartin which version of AixLib do I have to use? Will this work with development of AixLib?

What holds us back to merge this?

KremerMartin:
@PRemmen Currently the Branch [issue770_CollectiveLOM](https://github.com/RWTH-EBC/AixLib/tree/issue770_LOMCollectiveIssue) can be used. As I see it the merge needs to be reviewed. As well there is one last task in the AixLib-Issue to be done, of which I do not know the current status. After solving that, we could also merge the AixLib-Branch into development.

PRemmen:
there seem to be a lot merge conflicts, I'll try to resolve them :)

teaser/data/input/teaserjson_input.py
```diff
@@ -140,11 +139,23 @@ def load_teaser_json(path, project):
             ]
             tz.use_conditions.typical_width = zone_in["use_conditions"]["typical_width"]
             tz.use_conditions.with_heating = zone_in["use_conditions"]["with_heating"]
-            tz.use_conditions.with_ideal_thresholds = zone_in["use_conditions"]["with_ideal_thresholds"]
```
PRemmen: @PRemmen add this again, as it got lost with merge

teaser/data/output/modelicatemplate/AixLib/AixLib_Multizone
```diff
@@ -3,6 +3,7 @@ within ${bldg.parent.name}.${bldg.name};
 model ${bldg.name}
   "This is the simulation model of ${bldg.name} with traceable ID ${bldg.building_id}"
 
+% if bldg.internal_gains_mode==1 or bldg.internal_gains_mode==2:
```
PRemmen: In general I find it better to use strings to decide something like this instead of integers. BUt we can also do this as a new feature (and then for whole teaser)

Issue #574, opened 2019-03-26
Adjust version number of supported libraries


MichaMans:
Current build is failing due to a new library release of the supported libraries

This issue should be reopened if the test fails again

XucYin:
The unit tests failed because the Modelica **Buildings** Library has a new release [6.0.0](https://github.com/lbl-srg/modelica-buildings) but the version number number of it is 5.1.0 in TEASER (ibpsa.py).

PR #595, opened 2019-07-18
update version number of 'Buildings' in ibpsa.py

XucYin:
The current version of 'Buildings' is 6.0.0 #574


MichaMans:
@XucYin Perfect! I think the Travis Test fails because of to many api requests to github from the Travis Servers. So should be fine. Nevertheless i'll quickly have a look

XucYin:
@MichaMans Thanks! ^-^

XucYin:
@MichaMans OK. I will do the merge. The checks show success for Python 3.6.

PR #611, opened 2019-10-15
Issue599 import scriptfrom dev

MartinRaetz:
This PR is related to the Issue599, it demonstrates an algorithm for importing room based building data from an excel file into TEASER. 

There is only a new python example file and an example excel file for import. 
The requirement for numpy is also added to the setup.py.


MartinRaetz:
As far as I understand the CI has succeeded. Nevertheless, in the overview, it is shown pending. Is this an issue to my code or related to the CI?

teaser/examples/e9_building_data_import_from_excel.py
```diff
@@ -0,0 +1,629 @@
...
+import shutil
+
+
+def import_data(path=None, sheet_names=None):
+    """
+    Import data from the building data excel file and perform some
+    preprocessing for nan and empty cells.
+    If several sheets are imported, the data is concatenated to one dataframe
+
+    :param path: path to the excel file that should be imported
```
PRemmen: this is not the documentatin style guide we use in TEASER :( sorry, can you adjust that?

teaser/examples/e9_building_data_import_from_excel.py
```diff
@@ -0,0 +1,629 @@
...
+    #   have same set_temp_heat actual value: use_set_back = Check!
+    # 2. HeaterOn, CoolerOn, hHeat, lCool, etc. can be hard coded in the text
+    # file
+    #   "teaser / data / output / modelicatemplate / AixLib /
+    #   AixLib_ThermalZoneRecord_TwoElement"
+    #   actual changes: Check!
+
+    # Parameters to be set for each and every zone (#HardCodedInput)
+    # -----------------------------
+    OutWallTilt = 90
```
PRemmen: throughout the document could you please check for naming convention, e.g. as these are (hardcoded) but non global parameters it would be out_wall_tilt

teaser/examples/e9_building_data_import_from_excel.py
MartinRaetz: Thanks for noticing, the style should now match.

teaser/examples/e9_building_data_import_from_excel.py
MartinRaetz: Thanks for noticing, variable naming is now according to PEP8.

MartinRaetz:
I have implemented the requested changes.

PR #615, opened 2019-11-15
Add danish tabula buildings to setup

### LINKED ISSUE #614: Danish Tabula archetypes missing in setup.py - causes incomplete installation
This is related to #613 :

When installing TEASER via `pip install --upgrade git+https://github.com/RWTH-EBC/TEASER@development` the installation is not working. The cause is that `project.py` imports the Danish tabula buildings, but these files are missing from the installation via the command above. I think that [this answer on Stackoverflow](https://stackoverflow.com/a/10851054/3219539) points us in a good direction: `'teaser.logic.archetypebuildings.tabula.dk'` is missing in the setup file. I'll create a branch to fix this, check that everything works and create a PR to the development branch.

marcusfuchs:
Closes #614:

This is a quick and simple one: The Danish tabula buildings were missing in `setup.py`, causing pip installation directly from GitHub to fail. This PR adds the missing line.


MichaMans:
@marcusfuchs Thanks for the fix! This works.

We should maybe integrate a test for the pip github installation in the future.

marcusfuchs:
@PRemmen @MichaMans Thanks for the quick review. In #563, we agreed that I would just approve PRs and you would merge. So I fully approve and looking forward to this being merged ;)

Issue #620, opened 2019-12-20
[development] use conditions profile should accept single values


PRemmen:
## What is the problem?

- e. g. if I want to specify profiles with a single value (e. g. set temperature = 293.15 K) I always have to pass over a list

## Why do we want to solve it?

- this is annoying (especially for automatic manipulation of paramters)

## How do we want to solve it?

- just add a little check in the setter like:

```python
if not isinstance(value, list):
            value = [value]
```

PR #621, opened 2019-12-20
add check for profiles to be single values

PRemmen:
solves #620 (many changes because of nicer format with black)


PRemmen:
Pretty sure unit tests will run once #594 is merged :)

PRemmen:
ready to merge (if tests pass) - only the properties changed, the rest is only black code formatting @DaJansenGit

PR #631, opened 2020-01-21
Issue627 update version

### LINKED ISSUE #627: Update version number in teaser.logic.calculation
## What is the problem?

- A quite stupid mistake :( with the latest release we haven't updated the AixLib version numbers, which will result in non functional Modelica models

- to run modelica simulation

- increase version number and also increase version number of TEASER 0.9.1.

DaJansenGit:
closes #627 by raising version number of AixLib to v0.9.1. After this we should merge into master and release v0.7.2 to make current AixLib and TEASER release work together.


PRemmen:
I'd recommend to increase version number of TEASER in this PR, any objections? @DaJansenGit

DaJansenGit:
@PRemmen You are right!

Issue #543, opened 2018-10-25
Support of sunblinds in the thermal zone model


MichaMans:
## What is the problem?

- The advanced model of the aixlib thermal zone supports sunblinds, we should support that

## Why do we want to solve it?

- This enhances the simulation a lot

## How do we want to solve it?

- [x] Check maximum solar irradiation limit when sunblind closes in window class and material type
- [x] Check shading factor if sunblind is closed
- [ ] Integrate both factors in the AixLib Multizone Template

MartinRaetz:
I want to add my conclusion of the present state of the development regarding this sunblind issue:

There are unused parameter in the TypeBuildingElements Json files.
*"shading_max_irr"
*"Shading_g_total"

The WindowElement parameter "shading_max_irr" is not passed on in Teaser.
"Shading_g_total" will be converted to "g_sunblind" but is also not passed on anymore.

In the Modelica AixLib there is then a variable in the "PartialVDI6007" model named "sunblind" which is passed over by the connector and set by a constant source. This "sunblind" parameter reduces the "Equivalent long wave temperature for windows" by the corresponding sunblind factor. Put simply, if sunblind = 1 no long wave radiation penetrates the room, if sunblind = 0 the complete long wave radiation enters the room.

The "sunblind" parameter is statically set to zero, no other value from the database sheets influences it.
Excerpt from Modelicas ThermalZone Class:   
* Modelica.Blocks.Sources.Constant constSunblindWall[zoneParam.nOrientations](each k=0) "Sets sunblind signal to zero (open)";
* Modelica.Blocks.Sources.Constant constSunblindRoof[zoneParam.nOrientationsRoof]
(each k=0) "Sets sunblind signal to zero (open)";

## My suggestions:
1. Delete the values from the TypeBuildingElements Json´s to reduce the complexity and do not mislead the user.
2. Implement the full use of sunblinds. In that case it should be discussed, whether the use of sunblinds should be defined with the used window type (TypeBuildingElements) or with the respective UsageType (UseConditions), or even in another way.

@PRemmen proposed the following 3 step approach:
1. shading_max_irr in die Berechnungen aufnehmen (logic.buildingsobjects.calculation), damit der alle schwellenwerte für alle Fenster und zonen sammelt 
[...]

PR #634, opened 2020-01-22
Issue543 sunblinds

PRemmen:
This implements a very simple external shading into teaser, the changes in AixLib are for PR here: https://github.com/RWTH-EBC/AixLib/pull/861 - again if this makes more trouble than it helps other people I can also keep the changes local :) 

https://github.com/RWTH-EBC/TEASER/issues/543


teaser/logic/buildingobjects/calculation/four_element.py
```diff
@@ -329,8 +329,11 @@ class FourElement(object):
         facing the thermal zone and the ambient.
     weighted_g_value : float
         Area-weighted g-Value of all windows.
-    g_sunblind : list of floats
-        G-Value of all sunblinds of each window in a list
+    shading_max_irr : list of float [W/m2]
+        Threshold when sunblind becomes active for the whole zone
```
MartinRaetz, line 333:
```suggestion
        Threshold value above which the sunblind becomes active for the whole zone. Threshold regards to the incoming irradiation level with the window direction. This value does not account for heat flux due to the outside temperature.
```

teaser/logic/buildingobjects/calculation/four_element.py
```diff
@@ -329,8 +329,11 @@ class FourElement(object):
...
+    shading_g_total : list of float
+        Factor representing how much of the solar irradiation goes through
+        the sunblind
```
MartinRaetz, line 336:
```suggestion
        Factor representing how much of the actual solar irradiation goes through
        the sunblind and enters the window element, for the case, that the sunblind is activated.
```

teaser/logic/buildingobjects/calculation/one_element.py
```diff
@@ -183,8 +183,11 @@ class OneElement(object):
         facing the thermal zone and the ambient.
     weighted_g_value : float
         Area-weighted g-Value of all windows.
-    g_sunblind : list of floats
-        G-Value of all sunblinds of each window in a list
+    shading_max_irr : list of float [W/m2]
+        Threshold when sunblind becomes active for the whole zone
```
MartinRaetz, line 187: Change according to the description we will agree on.

teaser/logic/buildingobjects/calculation/three_element.py
```diff
@@ -258,8 +258,11 @@ class ThreeElement(object):
         facing the thermal zone and the ambient.
     weighted_g_value : float
         Area-weighted g-Value of all windows.
-    g_sunblind : list of floats
-        G-Value of all sunblinds of each window in a list
+    shading_max_irr : list of float [W/m2]
+        Threshold when sunblind becomes active for the whole zone
+    shading_g_total : list of float
```
MartinRaetz, line 263: Change according to the description we will agree on.

teaser/logic/buildingobjects/calculation/two_element.py
```diff
@@ -219,8 +219,11 @@ class TwoElement(object):
         facing the thermal zone and the ambient.
     weighted_g_value : float
         Area-weighted g-Value of all windows.
-    g_sunblind : list of floats
-        G-Value of all sunblinds of each window in a list
+    shading_max_irr : list of float [W/m2]
+        Threshold when sunblind becomes active for the whole zone
+    shading_g_total : list of float
```
MartinRaetz, line 224: Change according to the description we will agree on.

teaser/logic/buildingobjects/calculation/one_element.py
```diff
@@ -932,15 +932,15 @@ def _fill_zone_lists(self):
                 self.shading_g_total.append(
```
MartinRaetz, line 932: A few lines above, line 920, has this to be set to 1 there? Well, it probably doesn`t matter because there is no window.
And has shading_max_irr to be added there as well?

teaser/logic/buildingobjects/calculation/four_element.py
PRemmen: okay

teaser/logic/buildingobjects/calculation/one_element.py
PRemmen: I'll check the default values (i.e. no shading) on monday

PRemmen:
I hope I added all your suggestions. I'll need to fix the default value in AixLib (I think the default value is 0, which is not correct)

Issue #635, opened 2020-01-27
BuildingElements area and orientation setter not working with no parent set


MichaMans:
Currently, the BuildingElements ```area``` and ```orientation``` setter checks if the parent of the parent is set. Which leads to the problem, that the BuildingElement class can not be used Standalone. This needs to be revised.

MichaMans:
@PRemmen Do we need to classify this as a bug?

PR #636, opened 2020-01-27
#635 should revise the setters

MichaMans:
This should solve #635


MichaMans:
As coveralls is blocking the PR but it already commented, I'll merge with admin privileges

Issue #640, opened 2020-04-15
Typos and improvement of example file (Excel-Import)


MartinRaetz:
https://github.com/RWTH-EBC/TEASER/tree/issue640_typos_and_more

## What is the problem?
-Excel import is complicated and could be formulated more generally

-Air has typo at density = 1000x higher 
Actual density = 1000

-more typos

## Why do we want to solve it?

- make it right

## How do we want to solve it?

-apply some (not all) improvements to the example file for the excel import

- change air density to 1.225 kg/m³

MartinRaetz:
two_element.py

self.cool_load = -self.heat_load

cool_load is never calculated and automatically set to 0 so that the cooler never works even if present. I suggest to set cool_load to a very high number or at least the same as the heat_load?
(has to be done for all elements.py)

MartinRaetz:
ratio_conv_rad_persons: float 
        describes the ratio between convective and radiative heat transfer
        of the persons. Default values are derived from
        :cite:`VereinDeutscherIngenieure.2015c`.
        AixLib: Used in Zone record for internal gains
        Annex: Used for internal gains


Calculation: convective/radiative or the other way around?

- [ ] Info added to code

PR #642, opened 2020-04-15
Issue640 typos and more

MartinRaetz:
#640 

Typos and some changes at the excel import example script.

### DIFF
--- a/teaser/logic/buildingobjects/buildingsystems/buildingahu.py
+++ b/teaser/logic/buildingobjects/buildingsystems/buildingahu.py
@@ -73,6 +73,9 @@ class BuildingAHU(object):
         timeline of relative humidity requirements for AHU simulation
     v_flow_profile : [int]
         timeline of desired relative v_flow of the AHU simulation (0..1)
+        note: regularly this is just a dummy, since modelica only uses
+        the rel. occupancy of a zone to calculate the supply flow.
+        This can be changed with the AixLib parameter "WithProfile"
 
     """
 
--- a/teaser/logic/buildingobjects/buildingsystems/buildingahu.py
+++ b/teaser/logic/buildingobjects/buildingsystems/buildingahu.py
@@ -73,9 +73,11 @@ class BuildingAHU(object):
         timeline of relative humidity requirements for AHU simulation
     v_flow_profile : [int]
         timeline of desired relative v_flow of the AHU simulation (0..1)
-        note: regularly this is just a dummy, since modelica only uses
-        the rel. occupancy of a zone to calculate the supply flow.
-        This can be changed with the AixLib parameter "WithProfile"
+        Note: The AixLib parameter "WithProfile" determines whether the
+            (v_flow_profile combined with "min_ahu and max_ahu") or the
+            (persons_profile combined with "min_ahu and max_ahu")
+            is used for the AHU supply flow calculations.
+            Per default: (v_flow_profile combined with "min_ahu and max_ahu")
 
     """
 
--- a/teaser/logic/buildingobjects/useconditions.py
+++ b/teaser/logic/buildingobjects/useconditions.py
@@ -131,16 +131,22 @@ class UseConditions(object):
         Annex: Not used (see Annex examples)
     min_ahu: float [m3/(m2*h)]
         Zone specific minimum specific air flow supplied by the AHU.
-        Air flow varies lineary with the persons_profile,
-        person=0 -> min_ahu, person=1 -> max_ahu.
         AixLib: Used on Multizone level for central AHU to determine total
-        volume flow of all zones.
+        volume flow of each zone.
+            Note: The AixLib parameter "WithProfile" determines whether the
+            (v_flow_profile combined with "min_ahu and max_ahu") or the
+            (persons_profile combined with "min_ahu and max_ahu")
+            is used for the AHU supply flow calculations.
+            Per default: (v_flow_profile combined with "min_ahu and max_ahu")
     max_ahu : float [m3/(m2*h)]
         Zone specific maximum specific air flow supplied by the AHU.
-        Air flow varies lineary with the persons_profile,
-        person=0 -> min_ahu, person=1 -> max_ahu.
         AixLib: Used on Multizone level for central AHU to determine total
-        volume flow of all zones.
+        volume flow of each zone.
+            Note: The AixLib parameter "WithProfile" determines whether the
+            (v_flow_profile combined with "min_ahu and max_ahu") or the
+            (persons_profile combined with "min_ahu and max_ahu")
+            is used for the AHU supply flow calculations.
+            Per default: (v_flow_profile combined with "min_ahu and max_ahu")
     with_ahu : boolean
         Zone is connected to central air handling unit or not
         AixLib: Used on Multizone level for central AHU.
--- a/teaser/logic/buildingobjects/useconditions.py
+++ b/teaser/logic/buildingobjects/useconditions.py
@@ -170,6 +170,7 @@ class UseConditions(object):
     schedules: pandas.DataFrame
         All time dependent boundary attributes in one pandas DataFrame, used
         for export (one year in hourly timestep.)
+        Note: python attribute, not customizable by user (derived from Json)
 
     """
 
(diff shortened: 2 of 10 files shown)


PRemmen:
I'll have a look into it next monday

teaser/logic/buildingobjects/useconditions.py
```diff
@@ -52,13 +52,11 @@ class UseConditions(object):
         Degree Days and Precipitation in Europe—analysis for the
         CELECT-project" by Benestad, 2008.
     heating_profile : list [K]
-        Heating setpoint for a day or similar. You can set a list of any
+        Heating setpoint of AHU for a day or similar. You can set a list of any
```
PRemmen, line 55: this is not correct, this heating_profile is meant for the static ideal heating

teaser/logic/buildingobjects/useconditions.py
```diff
@@ -52,13 +52,11 @@ class UseConditions(object):
...
         length, TEASER will multiplicate this list for one whole year.
     cooling_profile : list [K]
-        Cooling setpoint for a day or similar. You can set a list of any
+        Cooling setpoint of AHU for a day or similar. You can set a list of any
```
PRemmen, line 58: this is not correct, this cooling_profile is meant for the static ideal cooling

teaser/logic/buildingobjects/useconditions.py
MartinRaetz: Thank you for clarifying! Now I wonder, why we can set the ideal heating threshold in 3 different ways: T_threshold_heating, heating_profile, and eventually schedule?

Does T_threshold_heating regard to the outside temperature? And marks a general shut-down of the component unregarding the zone temperature?

Do you know how the attribute schedule works? I think it could need a few more explaining words as well. If you comment some here, I can add them.

PRemmen: > Does T_threshold_heating regard to the outside temperature? And marks a general shut-down of the component unregarding the zone temperature?

Correct!

> Do you know how the attribute schedule works? I think it could need a few more explaining words as well. If you comment some here, I can add them. 

Attribute schedule is private and should not be used directly (thus has no documentation) There are only two attributes for indoor temperature `heating_profile`, `cooling_profile`

MartinRaetz: Thank you again for clarifying!

Please correct me, but in that case, there is no attribute to define the supply air temperature of the AHU within the JSON. So the supply temperature has to be set via central_ahu.temperature_profile within python. Which makes sense, because this is a building-related parameter and not a zone related one.

Do you know how the use_conditions.min_ahu/max_ahu relate with central_ahu.v_flow_profile?

As I could find out there is a parameter "withprofile" in modelica which defines whether the SUP flow is calculated via:

- withprofile = true -> via central_ahu.profile and min_ahu/max_ahu

- withprofiile = false -> rel. occupancy and min_ahu/max_ahu

But I was not able to find where this parameter "withprofile" is set.

Auszug Modelica: 
block AixLib.ThermalZones.ReducedOrder.Multizone.BaseClasses.AirFlowRateSum 
  "Air flow rate for AHU and unit conversion"
  extends Modelica.Blocks.Icons.Block;

  parameter Integer dimension "Number of Zones";
  parameter Boolean withProfile=false 
    "Profile or occupancy as control value for AHU";

After I understood the behavior, I will extend the documentation.

MartinRaetz: I cannot find "WithProfile" in teaser\data\output\modelicatemplate\AixLib\AixLib_Multizone
I cannot find it somewhere else in AixLib where it could be overwritten.
I cannot find a parameter in Teaser that could define this "WithProfile"

At my current state of research I suspect that calculations are always done via the rel. occupancy and the AHU profile has no effect. (Except someone changes the "WithProfile" value within Modelica)

MartinRaetz:
In the last push I provide a solution for the case that the review comments are correct.

teaser/logic/buildingobjects/useconditions.py
PRemmen: I think this is due to the fact that withProfile is by default=true in the Model `AixLib.ThermalZones.ReducedOrder.Multizone.MultizoneEquipped` that is used in TEASER export. thus i believe the v_flow_profile and min/max_ahu correlate directly in a sense that the maximum air volume supplied to the zone is:

```Modelica
((zoneParam.minAHU + (zoneParam.maxAHU -
    zoneParam.minAHU) * profile) .* zoneParam.AZone)

Comment (PRemmen) [reply teaser/logic/buildingobjects/useconditions.py]: > Please correct me, but in that case, there is no attribute to define the supply air temperature of the AHU within the JSON. So the supply temperature has to be set via central_ahu.temperature_profile within python. Which makes sense, because this is a building-related parameter and not a zone related one.

that is correct

> After I understood the behavior, I will extend the documentation.

That discussion and the extension help a lot :) thanks

Comment (MartinRaetz) [reply teaser/logic/buildingobjects/useconditions.py]: Thanks for evaluating my thoughts!

I think the default withProfile value is false!? (current development of AixLib)
So it is always occupancy controlled.

I would open an issue and link this discussion to it.
For this PR I would leave it with the docstring extension.

**Comment by MartinRaetz** [comment]
**All changes done, and confirmed by @PRemmen** 

Ready to merge.

File: teaser/logic/buildingobjects/useconditions.py (hunk 1 above)
Comment (PRemmen) [reply teaser/logic/buildingobjects/useconditions.py]: > I think the default withProfile value is false!? (current development of AixLib)
So it is always occupancy controlled.

I don't think so: https://github.com/RWTH-EBC/AixLib/blob/master/AixLib/ThermalZones/ReducedOrder/Multizone/MultizoneEquipped.mo#L133 (master - but dev should be the same)

File: teaser/logic/buildingobjects/useconditions.py [hunk 3]
```diff
@@ -123,21 +125,28 @@ class UseConditions(object):
         AixLib: Used for internal gains profile on top-level
         Annex: Not used (see Annex examples)
     min_ahu: float [m3/(m2*h)]
-        Zone specific minimum specific air flow supplied by the AHU
+        Zone specific minimum specific air flow supplied by the AHU.
+        Air flow varies lineary with the persons_profile,
+        person=0 -> min_ahu, person=1 -> max_ahu.
```
PRemmen, line 135: please double check with AixLib and export

teaser/logic/buildingobjects/useconditions.py
```diff
@@ -123,21 +125,28 @@ class UseConditions(object):
...
         AixLib: Used on Multizone level for central AHU to determine total
         volume flow of all zones.
     max_ahu : float [m3/(m2*h)]
-        Zone specific maximum specific air flow supplied by the AHU
+        Zone specific maximum specific air flow supplied by the AHU.
+        Air flow varies lineary with the persons_profile,
+        person=0 -> min_ahu, person=1 -> max_ahu.
```
PRemmen, line 141: please double check with AixLib and export

teaser/logic/buildingobjects/buildingsystems/buildingahu.py
```diff
@@ -73,6 +73,9 @@ class is very AixLib specific and holds some parameters that are only
         timeline of relative humidity requirements for AHU simulation
     v_flow_profile : [int]
         timeline of desired relative v_flow of the AHU simulation (0..1)
+        note: regularly this is just a dummy, since modelica only uses
+        the rel. occupancy of a zone to calculate the supply flow.
+        This can be changed with the AixLib parameter "WithProfile"
```
PRemmen, line 78: please lets double check

teaser/logic/buildingobjects/useconditions.py
MartinRaetz: I was wrong. I checked the base class, multizoneEquipped overwrites this value. 

I will change everything accordingly.

teaser/logic/buildingobjects/buildingsystems/buildingahu.py
MartinRaetz: Please see my latest changes, and confirm after you´ve double-checked it.

teaser/logic/buildingobjects/useconditions.py
MartinRaetz: Please see my latest changes, and confirm after you´ve double-checked it.

teaser/logic/buildingobjects/useconditions.py
MartinRaetz: Please see my latest changes, and confirm after you´ve double-checked it.

MartinRaetz:
Last commit should be named: "change with_ideal_threshold docstring"

PR #644, opened 2020-04-18
fix wrong assignment of humidity

### LINKED ISSUE #643: Typo in setter of max_relative_humidity_profile
# What is the problem?

- in the ahu the setter of max_relative_humidity_profile actually sets min_relative_humidity_profile 

- to set min and max bounds for humidity

- correct typo

PRemmen:
This fixes #643


MartinRaetz:
Thank you, @PRemmen!

Issue #612, opened 2019-11-05
Add CO2 balance


MartinRaetz:
## What is the problem?

The AixLib ROM shall be extended with a CO2 balance.
This is related to the AixLib Issue: https://github.com/RWTH-EBC/AixLib/issues/797 

[Branch: issue612_co2Model](https://github.com/RWTH-EBC/TEASER/tree/issue612_co2Model)
## Why do we want to solve it?

Teaser has to be adjusted, to provide the possibility to select this CO2 balance.
Otherwise, it is not compatible to the ROM or does not use the full spectrum of functionalities of the ROM. 

## How do we want to solve it?

- add the new AixLib ROM moisture/co2 model to the export files
- update documentation

MartinRaetz:
To be renamed and documented:

- [ ] use_moisture_balance to use_moisture_co2_balance

Necessary parameter in the AixLib export datasheets:
- [ ] activityDegree

PR #661, opened 2020-10-29
Issue612 co2 model

MartinRaetz:
This solves: https://github.com/RWTH-EBC/TEASER/issues/612


MartinRaetz:
I actually think the lines 72 - 77 in teaser/data/output/modelicatemplate/AixLib/AixLib_Multizone could be deleted completely. The thermal zone is automatically set in the modelica multizone models.

MartinRaetz:
@katharinabr implemented a reduced number of thermal zones in the AixLib (https://github.com/RWTH-EBC/AixLib/pull/1052). The use of Co2-balance and Moisture-balance is now optional via: 
use_C_flow = true,
use_moisture_balance = true,

Should we think about abandoning the internalgainsmode = 3 and set all co2 and moisture balance directly?

MartinRaetz:
There seems to be an issue with the Travis-ci. Even after 4 days and retriggering it is still pending.

Nevertheless, the Travis-ci should not be required anymore, since we have coveralls?

@DaJansenGit, I think it would be okay to perform a "force merge" if you approve my changes. Otherwise, let me know and I will try to get more information about Travis and Coveralls.

DaJansenGit (approved):
everything fine! :)

Issue #662, opened 2020-11-19
Bug in Tabula building area calcuation


marcusfuchs:
By default the  `SingleFamilyHouse` classes in `teaser/logic/archetypebuildings/tabula/de/singlefamilyhouse.py` and `teaser/logic/archetypebuildings/tabula/dk/singlefamilyhouse.py` contain 1 single zone. For a special use case, we want to add a second zone to this archetype. This exposed a bug in the area calculations performed in `generate_archetype()`. All areas are calculated similar to:

```python
outer_wall.area = (
    self.facade_estimation_factors[self.building_age_group]["ow1"]
    * type_bldg_area
) / len(self._outer_wall_names_1)
```

For 1 zone, this leads to correct results. But for more than 1 zone the total outside area of the building is not distributed between the zones. Instead, each zone gets the full outside area assigned to it, so that 2 zones result in twice the outside area compared to a 1-zone model for the same building. This is because the equations use `type_bldg_area` when in fact they should use `zone.area` instead. The following leads to correct results:

```python
outer_wall.area = (
    self.facade_estimation_factors[self.building_age_group]["ow1"]
    * zone.area
) / len(self._outer_wall_names_1)
```

As IMO this is a bug, I will branch of the master branch and submit a PR back to master.

PR #666, opened 2020-12-01
Fix area calculation by using correct base area

MichaMans:
Thanks to @marcusfuchs this implements the correct area calculation, see #662 #663


MichaMans:
@PRemmen unfortunately I needed to transfer this to an internal branch. I updated the version number. I would merge this after you approved, checks passing, seems a good solution

PRemmen (approved):

MichaMans:
this closes #662

Issue #665, opened 2020-11-24
Common base class for NonResidential and Residential archetypes?


marcusfuchs:
From looking at `teaser/logic/archetypebuildings/nonresidential.py` and `teaser/logic/archetypebuildings/residential.py` it seems like these two base class definitions for non-residential and residential buildings do not have any functional differences. As there seems to not have been any case where a distinction is necessary since these files were created in 2015, I wanted to suggest to replace these two files by a common `Archetype` base class definition. If there are no objections I could create a MR.

marcusfuchs:
As there were no objections, I implemented this, will create a PR.

PR #668, opened 2020-12-03
Use common base class for all Archetypes

marcusfuchs:
For #665. From looking at the code, there have been separate base classes for `Residential` and `NonResidential` archetypes since 2015, but there was never any need for these to differ. Therefore, I suggest to simplify things into one common `Archetype` base class. I think that this makes for a clearer starting point to create new archetypes and an easier foundation to understand the currently available ones.

As there were no differences between the two previous base classes, this does not lead to any functional changes.


PRemmen (approved):

PRemmen:
I'll transfer it to a teaser branch to get CI running and then merge if there are not tests failing

Issue #669, opened 2020-12-04
Enable regression testing of TEASER output with BuildingsPy


marcusfuchs:
For quality control, I think it would be really useful if we could easily run all models generated from a TEASER project as regression tests. With BuildingsPy this can be achieved fairly easy. As TEASER outputs a Modelica package, all that is need for this is to add the following things to the output:

- Add a script e.g. `/Resources/Scripts/runUnitTests.py` similar to https://github.com/RWTH-EBC/AixLib/blob/development/bin/02_CITests/UnitTests/runUnitTests.py
- Add annotations to each building model that enable using the model as a regression test
- Add MOS scripts to run the regression tests as in https://github.com/RWTH-EBC/AixLib/tree/development/AixLib/Resources/Scripts/Dymola

This way, BuildingsPy can directly run the simulation of all buildings as a regression test and generate reference results. Then users can always verify whether the models still give the same results.

I can suggest an implementation for this, but feedback and help is of course very welcome

MichaMans:
@marcusfuchs I like this addition! I may check an automated testing in the CI afterwards. 👍

marcusfuchs:
@MichaMans Yes, I think this would be perfect. We already have implemented something similar as would be needed for TEASER, so once I get around to work on this I think the effort required will not be too large. For automated testing in your CI afterwards, I also think that https://github.com/lbl-srg/BuildingsPy/issues/393 will be helpful so you do not have to copy the reference results into the output directory.

marcusfuchs:
@MichaMans Starting to use this I noticed one error: I forgot to add the `modelica_test_script` template to `setup.py`, so when installing TEASER from GitHub with `pip install --upgrade git+https://github.com/RWTH-EBC/TEASER@development` the file is missing. I'll create a PR to fix this.

marcusfuchs:
@MichaMans Thanks again, this is all working as intended, so I am closing this.

PR #670, opened 2020-12-10
Enable regression testing of model output with BuildingsPy

marcusfuchs:
For #669:

The Modelica output generated by TEASER now includes all scripts and annotations to easily run regression tests on all the building models. From the top-level of the model output, simply run `python Resources\Scripts\runUnitTests.py` and everything should work. For testing against existing results e.g. in CI pipelines, `Project.dir_reference_results` also copies existing reference results into the TEASER output.

@MichaMans what do you think about the next steps for this? From my perspective, this has everything we would need for our internal CI pipelines.


MichaMans:
@marcusfuchs this should help implementing it in the CI for auto test the simulation. I will transfer that to an internal branch. How exactly is the use of the copy of the reference results working or how should we use that for the test process?

MichaMans:
now in #672

PR #672, opened 2020-12-28
Issue669 buildingspy

MichaMans:
Transfer from #670


marcusfuchs:
@MichaMans Thanks, I'm happy to see that you pick this up! Regarding your question 

>How exactly is the use of the copy of the reference results working or how should we use that for the test process?

Until it may be possible to specify a custom path for the reference results with https://github.com/lbl-srg/BuildingsPy/pull/394 in BuildingsPy, the copying allows this workflow:

- You can place your reference results in any folder within TEASER, e.g.  `TEASER/tests/ReferenceResults`
- Setting `prj.dir_reference_results = ` to this path before model export, e.g. in a test, will copy the reference results into the output model library at the right location `Resources/ReferenceResults/Dymola/` for BuildingsPy to find them
- Thus, if you point `prj.dir_reference_results` to any folder of reference results before model export, you can directly run the `runUnitTests.py` script also copied to output model library to simulate and compare for your tests.

MichaMans:
@marcusfuchs ahh ok. If I understand correctly that is meant then for the reference results (which I need to create locally in the first place) to place them somewhere in teaser (e.g. teaser/examples/examplefiles/ArchetypeExampleReferenceResults) and then point to that directory before export.

marcusfuchs:
@MichaMans Yes, exactly. The open PR in BuildingsPy will hopefully help to make things easier, but I think you would probably often want to create the reference results locally anyway before you decide that they will be the reference for CI.

MichaMans:
@marcusfuchs ok perfect. One more thing as the runUnitTest.py is leading to failing tests. is the OpenModelicaUnitTest a thing for us? In my point of view not. So I am deleting that part. If you have any objections we can add that again

marcusfuchs:
No feel free to delete that, we also do not need it.

MichaMans:
@marcusfuchs I am currently making some steps towards the ci integration. Have you set this one up before? Am I able to load the AixLib via BuildingsPy to make the simulation of the generated models possible?

marcusfuchs:
@MichaMans I have set up similar setups, but not exactly with this TEASER implementation. To be honest, I was waiting for this to be merged into development in order to avoid effort setting this up from my branch and then again working on it to switch back to the develoment branch. But in our applications, we would let AixLib be loaded automatically by putting it on the MODELICAPATH of a GitLab runner machine's environment variables.

MichaMans:
@marcusfuchs Thought so, do you know if this works also with a dockered dymola container oder with linux?

marcusfuchs:
@MichaMans Sorry, no experience with either of your two cases

MichaMans:
@marcusfuchs ok. Then I suggest that I add the reference results for the example e2 and in e2 I'll add the path to the directory. Which seems to work as expected. After that we can merge that integration and add the actual CI testing in another step.

MichaMans:
@marcusfuchs so if you are fine with what I added, I'll merge this

marcusfuchs:
@MichaMans Thanks, I'd be very happy about a merge. Then I can also use this in our application and can share our experiences if that helps.

MichaMans:
@marcusfuchs please see if all in #669 is resolved. Then you can close that

marcusfuchs:
@MichaMans Thanks, I'll close after confirming that everything works.

PR #673, opened 2020-12-30
Issue669 setup

marcusfuchs:
For #669 : 

We added a new template `modelica_test_script`. This needs to be added to `package_data` in `setup.py`, otherwise the file is missing and TEASER will not work when installed directly from GitHub via pip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
             "package_order",
             "conversion",
             "modelica_language",
+            "modelica_test_script",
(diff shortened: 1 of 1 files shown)


marcusfuchs:
@MichaMans I verified that this change does solve the encountered problem in our CI system.

MichaMans:
@marcusfuchs thanks, I transferred that to #674 which I'll merge if checks passing

PR #687, opened 2021-04-19
solve export problem

### LINKED ISSUE #686: export function prevents folder deletion
## What is the problem?
**Changed working directory:**
Calling the export function prevents to delete the export folder because the utilities.create_path() function changes the current working directory.
This leads to problem on windows (maybe also on linux, which I didn't test), when the project folder should be deleted by hand or program after running an export without changing the python working directory afterwards.
**File exporting**
Also the opening of the files is not safe, because no `with` statement is used. This could also lead to problems.

* Make TEASER better usable in combination with other tools/programs

**Changed working directory:**
Lines to remove:
* https://github.com/RWTH-EBC/TEASER/blob/development/teaser/logic/utilities.py#L36
* https://github.com/RWTH-EBC/TEASER/blob/development/teaser/logic/utilities.py#L42
**File exporting**

change lines like this:
```
out_file = open(utilities.get_full_path
                        (os.path.join(bldg_path, bldg.name + ".mo")), 'w')
```
to 
```
with open(utilities.get_full_path
                        (os.path.join(bldg_path, bldg.name + ".mo")), 'w') as out_file:
```

DaJansenGit:
closes #686

### DIFF
--- a/teaser/data/output/aixlib_output.py
+++ b/teaser/data/output/aixlib_output.py
@@ -140,15 +140,14 @@ def export_multizone(buildings, prj, path=None):
                                                "number of the building in "
                                                "the project list.")
                 bldg.building_id = i
+        with open(utilities.get_full_path(
+                os.path.join(bldg_path, bldg.name + ".mo")), 'w') as out_file:
 
-        out_file = open(utilities.get_full_path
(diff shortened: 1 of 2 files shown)


MichaMans (approved):
@DaJansenGit Looks good, thanks for the implementation

PR #692, opened 2021-07-23
Pypi actions

MichaMans:
This action should autodeploy teaser to pypi if a new tag is created. @MartinRaetz could you approve?


MartinRaetz (approved):
Thank you @MichaMans!

I am not familiar with those actions. Your code looks reasonable and your expertise makes this PR trustworthy enough to give you a "blind" approve!

PR #706, opened 2022-01-05
fix: integrate unique zone names to solve #705

### LINKED ISSUE #705: non unique thermalzone names lead to false internal gains schedules
## What is the problem?

- currently the setter function for `thermalzone.name` doesn't check if the name is unique (same name for other zone) ([code](https://github.com/RWTH-EBC/TEASER/blob/master/teaser/logic/buildingobjects/thermalzone.py#L478-L487))
- the pandas df which is used to generate the `InternalGains_*.txt` file uses the thermalzone names for column naming
- as pandas df by default only accepts unique column names (for good reason), existing columns with same name will be overwritten.
- this leads to an array where the number of columns < number of zones and therefore to false model export.

- even if double names for thermalzones shouldn't be used in TEASER, especially for automatic generated models this can occure
- it doesn't lead to an error in TEASER and the model exports fine but can't be simulted as given size in the header of the exported txt and real size doesn't match. Modelica error:
```
The stack of functions is:
Modelica.Blocks.Types.ExternalCombiTimeTable
Modelica.Blocks.Types.ExternalCombiTimeTable((if tableInternalGains.tableOnFile then tableInternalGains.tableName else "NoName"), (if tableInternalGains.tableOnFile and tableInternalGains.fileName <> "NoName" and  not Modelica.Utilities.Strings.isEmpty(tableInternalGains.fileName) then tableInternalGains.fileName else "NoName"), tableInternalGains.table, tableInternalGains.startTime/tableIntern...

FixInitials:Init: Integrator failed to start model.
```

- add a check in the setter function for `thermalzone.name` to manipulate the name to be unique to prevent the described behaviour

DaJansenGit:
closes #705

### DIFF
--- a/teaser/logic/buildingobjects/thermalzone.py
+++ b/teaser/logic/buildingobjects/thermalzone.py
@@ -480,14 +480,11 @@ class ThermalZone(object):
     @name.setter
     def name(self, value):
         regex = re.compile('[^a-zA-z0-9]')
-        # name = None
         if isinstance(value, str):
-            # self._name = regex.sub('', value)
             name = regex.sub('', value)
         else:
             try:
                 value = str(value)
-                # self._name = regex.sub('', value)
                 name = regex.sub('', value)
 
             except ValueError:
--- a/teaser/logic/buildingobjects/thermalzone.py
+++ b/teaser/logic/buildingobjects/thermalzone.py
@@ -4,6 +4,8 @@
 """This module includes the ThermalZone class
 """
 from __future__ import division
+
+import itertools
 import random
 import re
 import warnings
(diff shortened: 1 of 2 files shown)


teaser/logic/buildingobjects/thermalzone.py
```diff
@@ -477,18 +479,30 @@ def name(self):
 
     @name.setter
     def name(self, value):
+        regex = re.compile('[^a-zA-z0-9]')
+        # name = None
         if isinstance(value, str):
-            regex = re.compile('[^a-zA-z0-9]')
-            self._name = regex.sub('', value)
+            # self._name = regex.sub('', value)
```
MartinRaetz, line 485: Can we delete the comments?

tests/test_data.py
```diff
@@ -3156,3 +3156,21 @@ def test_export_bldg_threshold(self):
...
+        tz1 = ThermalZone(parent=bld)
+        tz1.name = "living"
+
+        tz2 = ThermalZone(parent=bld)
+        tz2.name = "kitchen"
+
+        tz3 = ThermalZone(parent=bld)
+        tz3.name = "living"
+
+        assert tz3.name == "living_1"
```
MartinRaetz, line 3176: Not important, but maybe assert all three zones?
Who knows what might go wrong in future adaptions.

tz1.name == "living" 
tz2.name == "kitchen"

teaser/logic/buildingobjects/thermalzone.py
DaJansenGit: fixed with 9859fd0448812017ff63fbafdcd68d7dfea9fcd2

tests/test_data.py
DaJansenGit: added by f134f3f79a75b2eaefcd6263bc1deabf5aceeec7

MartinRaetz (approved):

Issue #711, opened 2022-04-14
Typo in the heating profile example (e8)


Dawac:
The example doesn't work. The heating profile does not change because the parameter has a typo "*.heating_profile_profile" instead of "*.heating_profile".

Dawac:
The bug had been solved.

PR #712, opened 2022-06-30
Corrected typo in Example 8

Dawac:
This solves a simple typo in Example 8, that caused a bug, described in #711 .

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,7 +7,7 @@ test3.6-subset:
     - pip install --only-binary=numpy,scipy numpy scipy
     - pip install pyxb
     - pip install pytest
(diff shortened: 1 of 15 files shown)


DaJansenGit:
thanks for removing this bug.
pipeline failed due to specification of pip install of old pygithub version in script section. I merged the current development where this is already fixed. Pipelines should succeed now. You can merge afterwards.

DaJansenGit (approved):
Thanks for removing this bug. 
Pipeline failed due to specification of pip install of old pygithub version in script section. I bumped the version to 1.55 which is the same as in dev. Pipeline should succeed now. You can merge afterwards. After merging this, we should apply the same fix to development.

PR #714, opened 2022-08-16
integrate parameters for dynamic heat load calc in modelica

### LINKED ISSUE #707: dynamic heat load calc in modelica
## What is the problem?

- Currently the heat load is calculated in TEASER but the needed variables are not passed into Modelica.

- As the heat load is a suitable value for system design and for parametric studies it can be useful to vary e.g. the nominal inside or outside temperature (other location of building) it would be useful to allow recalculation of the heatload directly in Modelica without creating a new TEASER model

- As the heat load calculation equation can be simplified to `A * (t_inside-t_outside) + B * (t_inside-t_ground)` we could simply include the parameters `A` and `B` into the TEASER export and allow recalculation inside modelica.

DaJansenGit:
closes #707 

should be merged together with [this AixLib PR](https://github.com/RWTH-EBC/AixLib/pull/1328)


FWuellhorst (approved):
Looks good, thank you!

FWuellhorst (approved):

PR #725, opened 2022-10-13
refactor retrofit wall function

### LINKED ISSUE #713: move retrofit_wall() into corresponding subclass
## What is the problem?

- currently the `retrofit_wall()` lives in wall.py base class but distinguishes by statements like `type(self).__name__ == 'OuterWall':` between the type of wall

- it might make sense to refactor this to prevent confusion and improve readability

1. rename the method `retrofit_wall()`method in wall.py into `retrofit_buildingelement()` and make it an abstract method.
2. move the corresponding part of the function into the subclasses 
e.g. the [OuterWall ](https://github.com/RWTH-EBC/TEASER/blob/master/teaser/logic/buildingobjects/buildingphysics/wall.py#L297-L316) part into [outerwall.py](https://github.com/RWTH-EBC/TEASER/blob/master/teaser/logic/buildingobjects/buildingphysics/outerwall.py) and overwrite the `retrofit_buildingelement()` method in each subclass
3. `insulate_wall()` could stay in wall.py, but should be renamed to `insulate_buildingelement()`

This doesn't change anything regarding the functionality, but might improve readability.
If the changes are wanted, I will implement them.

DaJansenGit:
closes #713 

This should improve readability and give some more feedback to the user if a retrofit was performed or not (e.g. in case of "too modern walls" with already good u values).

### DIFF
--- a/teaser/logic/buildingobjects/buildingphysics/groundfloor.py
+++ b/teaser/logic/buildingobjects/buildingphysics/groundfloor.py
@@ -77,6 +77,7 @@ class GroundFloor(OuterWall):
     c1_korr : float [J/K]
         corrected capacity C1,korr for building elements in the case of
         asymmetrical thermal load given in VDI 6007
+    calc_u: Required area-specific U-value in retrofit cases [W/K]
     ua_value : float [W/K]
         UA-Value of building element (Area times U-Value)
     r_inner_conv : float [K/W]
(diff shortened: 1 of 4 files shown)


KremerMartin (approved):
Maybe we could add a Docsting for the variable `calc_u`. But just a suggestion. Should be fine like this already.

PR #726, opened 2022-10-20
hotfix to not insulate when u values are same

### LINKED ISSUE #713: move retrofit_wall() into corresponding subclass

DaJansenGit:
minor bugfix to prevent layer of insulation with zero thickness which leads to incorrect RExt parameters in export

closes #713

### DIFF
--- a/teaser/logic/buildingobjects/buildingphysics/wall.py
+++ b/teaser/logic/buildingobjects/buildingphysics/wall.py
@@ -305,7 +305,7 @@ class Wall(BuildingElement):
     def set_insulation(self, material, calc_u, year_of_retrofit):
         """Sets the correct insulation thickness based on the given u-value"""
         if calc_u:
-            if self.u_value < calc_u:
+            if self.u_value <= calc_u:
                 warnings.warn(
                     f'No retrofit needed for {self.name} as u value '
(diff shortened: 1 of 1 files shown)


KremerMartin (approved):
Thanks for fixing. :)

Issue #683, opened 2021-04-08
Model Report


DaJansenGit:
> Just stopping by to let us know about some issue you encountered with TEASER? Please feel free to delete all of this text and tell us what you think!

> If you want to suggest larger changes and/or adding new features to TEASER, please make sure to briefly answer the following few questions. For more information on how to contribute to TEASER, see our [Wiki](https://github.com/RWTH-EBC/TEASER/wiki). Before submitting the issue, you can delete these first lines. Thanks for contributing!

## What is the problem?

- Currently when creating buildings with TEASER we have no report that gives us combined information about what the model holds. Information that might be usefull to verify that everything is correct could be:
- Net floor area of building
- Area of outerwalls (orientation wise)
- Area of windows (orientation wise)
- Area of Innerwalls
- Number of Thermalzones
- ...

## Why do we want to solve it?

- To get a better overview about the models we create.

## How do we want to solve it?

A first shot could be some simple html report that is porvided by the TEASER output.

PR #744, opened 2023-11-06
Improve model report

DaJansenGit:
Minor improvements to make the model export a bit more robust

If no `height_of_floors` is given (only mandatory for archetypes), a default value of 3 m is used. If this is the case this is also stated in the model report (see attached screenshot)

image


MartinRaetz:
Related to #683. Will this close the issue?

MartinRaetz (approved):
Thank you for the fix, looks well implemented.

PR #749, opened 2024-02-15
add __Dymola_selection implementation

### LINKED ISSUE #748: Add __Dymola_selection option to AixLib export
## What is the problem?
- Currently, TEASER exports always store all variables by default, one can use the Dymola output options to select to store only outputs, but this is not really useful as not all relevant data from the exported models are stored in outputs.
- Due to this result files might get really big especially for buildings with a lot of thermal zones.

- More granular selection of stored data in result files and reduction of file size for fast post-processing 

Use the __Dymola_selection option. 
Therefore I implemented an additional parameter `export_vars` in the `export_aixlib()` method. The parameter is a dictionary that specifies which parameters should be stored. Following you can find an example and the results of this. I also added some documentation in the example `e2_export_aixlib_models`. 

### Test with the following export selected:
```    export_vars = {
        "HeatingDemands": ["*multizone.PHeater*", "*multizone.PHeatAHU"],
        "CoolingDemands": ["*multizone.PCooler*", "*multizone.PCoolAHU"],
        "Temperatures": ["*multizone.TAir*", "*multizone.TRad*"]
    }
```

Standard: **8166 KB**
Reduced result file size: **429 KB**

Here one can see how the results are displayed:
image

If not providing the parameter `export_vars` to the `export_aixlib()` method, everything behaves like "normal".

DaJansenGit:
closes #748. Relevant information can be found in the issue!

### DIFF
--- a/teaser/project.py
+++ b/teaser/project.py
@@ -1034,7 +1034,8 @@ class Project(object):
         internal_id=None,
         path=None,
         use_postprocessing_calc=False,
-        report=False
+        report=False,
+        export_vars=None
     ):
         """Exports values to a record file for Modelica simulation
 
@@ -1059,6 +1060,11 @@ class Project(object):
         report: boolean
             if True a model report in form of a html and csv file will be
             created for the exported project.
+        export_vars: dict[str:list]
+            dict where key is a name for this variable selection and value is a
+             list of variables to export, wildcards can be used, multiple
+             variable selections are possible.
+
         """
 
         if building_model is not None or zone_model is not None or corG is not None:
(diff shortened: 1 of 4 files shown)


teaser/project.py
```diff
@@ -1059,6 +1060,11 @@ def export_aixlib(
         report: boolean
             if True a model report in form of a html and csv file will be
             created for the exported project.
+        export_vars: dict[str:list]
+            dict where key is a name for this variable selection and value is a
+             list of variables to export, wildcards can be used, multiple
+             variable selections are possible.
```
FWuellhorst, line 1066: Why the whitespace in the second and third line?

FWuellhorst: Link to claytex o.s.? 
Also, explicitly note that this will only work for Dymola?

FWuellhorst (approved):
Minor comments on docs, nice work

teaser/project.py
DaJansenGit: you are right! both solved with 7cd15517

FWuellhorst (approved):

PR #754, opened 2024-04-26
fix outer area calculation for visualization

DaJansenGit:
Small fix. The prior code lead depending on the structure of the model to wrong assignments of the orientation of the outer wall areas.


MartinRaetz (approved):
Without testing it myself, the proposed change with explanation makes sense and is tested by @DaJansenGit. Thank you for the PR @DaJansenGit!

PR #757, opened 2024-05-07
Issue745 add kfw archetypes

jonas0902:
closes #745

PR #758, opened 2024-05-07
Update development with changes from master

FWuellhorst:
The master and development are out of sync, which leads to issues in #757 . We should update them separately.


FWuellhorst:
@DaJansenGit : Most changes are for the new CI structure. However, some files are changed using some new features you implement at some point. Can you check if everything is compatible?

tests/test_data.py
```diff
@@ -3156,23 +3156,3 @@ def test_export_bldg_threshold(self):
...
-
-        tz2 = ThermalZone(parent=bld)
-        tz2.name = "kitchen"
-
-        tz3 = ThermalZone(parent=bld)
-        tz3.name = "living"
-
-        assert tz1.name == "living"
-        assert tz2.name == "kitchen"
-        assert tz3.name == "living_1"
```
DaJansenGit, line 3178: why did you delete this test?

DaJansenGit (commented):
Thanks for the good work! There is just one removed test I'm not sure about (see comment)

tests/test_data.py
FWuellhorst: Not intendet, this test does not exist in the master, maybe a merge conflict. It should be in there.

DaJansenGit: Ok, I commented the test now as the code that is needed for test is only in dev and not in this branch. Don't complety understand what went wrong here. But lets merge this into dev after the tests completed and then merge dev -> master. We need to uncomment this test then again by hand before the merge into master.

DaJansenGit (approved):

PR #763, opened 2024-05-28
753 add illuminance levels and lighting efficiencies in use conditions

### LINKED ISSUE #753: Add illuminance levels and lighting efficiencies in use-conditions
## What is the problem?

- At the moment, internal lighting gains are only based on the input of fixed electrical lighting power per zone [W/m²] in the use conditions.

- In standards for usage boundary conditions of buildings (e.g. DIN V 18599-10, SIA 2024) only the illuminance level [lx = lm/m²] for specific zones is defined. 
- The electrical lighting power can then be determined using the lighting efficiency [lm/W] which depends on the used lighting device (halogen, LED, etc.)

- Additional getter and setter methods are added in the useconditions.py class, so that the electrical lighting power per zone is determined by one of two options:
1) Fixed input of the electrical lighting power per zone in the use-conditions
2) Calculation of the electrical lighting power per zone by the input of an illuminance level and lighting efficienciy per zone in the use-conditions
- By adding a boolean variable, it will be possible to switch between these two options.
- Option 1) will be selected as default

HoeppJonatan:
closes #753


DaJansenGit:
We will merge #757 first and afterward make the required changes here (if needed) and merge this as well.

teaser/data/input/inputdata/UseConditions.json
DaJansenGit, line 1: please revert the line breaking changes on `persons` as these just mess with the hisotry and confuse

teaser/logic/buildingobjects/useconditions.py
```diff
@@ -244,8 +260,14 @@ def __init__(self, parent=None):
         self.machines = 7.0
         self.ratio_conv_rad_machines = 0.75
 
-        self.lighting_power = 15.9
+        self._use_maintained_illuminance = False    #Choose wether lighting power will be given by direct input or calculated by maintained illuminance and lighting_efficiency_lumen
```
DaJansenGit, line 263: `# Choose` add space here.

teaser/logic/buildingobjects/useconditions.py
```diff
@@ -779,3 +801,31 @@ def parent(self, value):
...
+    def lighting_power(self):
+        return self._lighting_power
+
+    @lighting_power.setter
+    def lighting_power(self, value):
+        if self.use_maintained_illuminance:
+            warnings.warn(
+                "Parameter 'use_maintained_illuminance' is 'True'!\n"
+                "Parameter 'lighting_power' will be overwritten and 'use_maintained_illuminance' will be set to 'False'.",
+                Warning
```
DaJansenGit, line 828: ? Remove `Warning` statement in line 828

DaJansenGit (changes requested):
Thanks for the implementation. Minor changes requested.

DaJansenGit:
GitLab is temporary down due to maintenance. If CI passes afterwards, we can merge.

PR #764, opened 2024-05-28
764 added exception

### LINKED ISSUE #762: Add
## What is the problem?
Method `process_export_vars` in `project.py` does not check if given items of dict `export_vars` are an instance of a list.
- ...

Returned string `export_vars_str` of method `process_export_vars` would lead to incorrect output variables in the dymola result file
- ...

Add code in the method `process_export_vars` to check, if all items of given dict `export_vars` are an instance of a list.
Raise error, when an item is not an instance of a list.
- ...

HoeppJonatan:
closes #762


DaJansenGit (approved):

DaJansenGit:
Thanks for the PR @HoeppJonatan :)

PR #773, opened 2024-10-02
fix bug with `with_ahu` setter method

### LINKED ISSUE #772: with_ahu setter method is incorrect
The current setter method of `with_ahu` of the Building class looks like this:
```python
    @with_ahu.setter
    def with_ahu(self, value):

        if value is True and self.central_ahu is None:
            self.central_ahu = BuildingAHU(self)
            self._with_ahu = True
        elif value is False and self.central_ahu:
            self.central_ahu = None
            self._with_ahu = False
```

The problem occurs in the following scenario:
* somebody already created a `BuildingAHU()` object from the outside and added it to Building.central_ahu, but `with_ahu`  is still on its default (False)
* now form the following statement is called: `Building.with_ahu = True`
* Now the first if clause is False because `self.central_ahu` is not None and `with_ahu` stays False.

DaJansenGit:
closes #772

### DIFF
--- a/teaser/logic/buildingobjects/building.py
+++ b/teaser/logic/buildingobjects/building.py
@@ -668,6 +668,8 @@ class Building(object):
         if value is True and self.central_ahu is None:
             self.central_ahu = BuildingAHU(self)
             self._with_ahu = True
+        elif value and self.central_ahu and self._with_ahu is False:
+            self._with_ahu = True
         elif value is False and self.central_ahu:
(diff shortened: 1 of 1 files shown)


FWuellhorst (approved):

DaJansenGit:
The test for python 3.7 fails, but this is also the case for current development, I opened an issue (#776) but will merge this PR now.

PR #775, opened 2024-10-04
774 retrofit combos

FWuellhorst:
@Nico-yf  @larissakuehn 

Based on the scripts you provided me to retrofit single building elements, I implemented this feature for TEASER. Could one or both of you check if the functions work for you in your lib and review the code? Be aware that you have to switch to the latest TEASER version to check.

Regarding naming of the buildings, I followed your approach,  @Nico-yf. We could add an option so that users may define the names based on a callable? E.g. :
```
def get_retrofit_name(element_retrofit_stats: dict):
    retrofit_name = ''.join(
            f"{element[:2]}{retrofit_dict[retrofit_option]}"
            for element, retrofit_option in element_retrofit_stats.items()
        )
    return retrofit_name
```

### DIFF
--- a/teaser/__init__.py
+++ b/teaser/__init__.py
@@ -7,7 +7,7 @@ Tool for Energy Analysis and Simulation for Efficient Retrofit
 import sys
 import os
 
-__version__ = "1.0.0"
+__version__ = "1.0.1"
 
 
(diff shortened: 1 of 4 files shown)


Nico-yf (approved):
Function works for me.  thanks for implementing!

Issue #679, opened 2021-02-25
Custom template paths for Modelica output


PGorzalka:
## What is the problem?

- The current implementation restricts templates to files located in pre-defined paths within TEASER's folder structure

## Why do we want to solve it?

- For custom applications, it would be nice to switch between different Multizone and ThermalZone templates

## How do we want to solve it?

- Add an optional parameter to the export_aixlib function that specifies a path to a template on the local hard drive

## Additional context

This issue is related to a case-specific application of AixLib (ROM) and TEASER which also includes issue [#1080](https://github.com/RWTH-EBC/AixLib/issues/1080) in AixLib (others may follow after email consultations)

PGorzalka:
Discussion with @DaJansenGit (see [here](https://github.com/RWTH-EBC/AixLib/issues/1080#issuecomment-1460022322)) lead to the envisaged contributions of
* non-constant soil temperature
* interzonal heat exchange (TEASER and AixLib)
* custom template paths for Modelica output in TEASER as described in this issue
* a new possibility to estimate interior wall area (relates to #474 )

PGorzalka:
solved by PR #786

PR #785, opened 2024-11-17
Issue679 interzonal elements, soil temperature, and interior geometry estimation

PGorzalka:
Finally, I was able to finish the work on issue #679 
@DaJansenGit thanks for your support!

Changes mainly encompass the work mentioned in [this paper](https://doi.org/10.3384/ecp204577):
- building up on [changes already implemented via IBPSA merge](https://github.com/ibpsa/modelica-ibpsa/issues/1744), some adaptations for non-constant soil temperature
  - Soil temperature can now be a constant, a sine, or specified in an additional input file
  - Project parameter defines what is chosen
- Interzonal elements (that is, building elements that connect two zones)
  - New building element object types `InterzonalWall`, `InterzonalCeiling` and `InterzonalFloor`
  - Due to current convention, building elements have only one parent thermal zone. This is why an interzonal element has to be created twice (once for each zone it borders) and has subsequently two representations in TEASER
  - Export to AixLib using the new FiveElements RC model there, making sure that only one of the two representations is actually considered in simulation.
  - If exported to an AixLib model with less elements, interzonal building elements will be lumped into interior or exterior walls depending on the zones' properties
  - Option to load layers in reversed order to make sure borders are correctly represented on both sides
  - Possibility to place insulation layer not necessarily on the outside (making sure that it is placed correctly for both representations of the same interzonal element)
- New method(s) to estimate interior geometry
  -  Introduced as string parameter of a Building object
  - `'typical_minus_outer'` sets length of inner walls = 2 * typical length * height of floors + 2 * typical width * height of floors - length of outer or interzonal walls
  - `'typical_minus_outer_extended'` like `'typical_minus_outer'`, but also considers that
     a) a non-complete "average room" reduces its circumference proportional to the square root of the area
     b) rooftops, windows and ground floors (= walls with border to soil) may have a vertical share
  - Number and height of floors can now be set as zone parameters
- Included everything to JSON input/output
- Example 11 covering all three (soil temperature, interzonals, new interior mass estimation) and creates ASHRAE140 test case 960
- Tests covering all three (soil temperature, interzonals, new interior mass estimation). Note: I did not change the old test functions, although the GitHub diff says I did. I just duplicated and adapted them.

Additional minor stuff:
- Custom multizone template, so boundary conditions can be adapted more easily
- two missing `each`s in `AixLib.Fluid.Pools.BaseClasses.AirFlowMoistureToROM` caused OM to fail -> fixed
- weather data path can be specified starting with `modelica:` to reference to existing Modelica resources
- Some fixes to buildingelement.py to make it more resilient against minor bugs
- Corrected some docstring typos
- Heavily reduced length of code in RC parameter calculation. Adaptations there were necessary due to the reversed layers of interzonal elements, but it was basically impossible to debug before.

Missing:
- Test Case 960 was not actually tested against standard results because the weather file provided in AixLib (AixLib\Resources\weatherdata\ASHRAE140.mos) does not contain the weather data provided by the standard. 
- Reference the paper in Readme?
- version count, also in example json inputs
- Project docstring that recommends `For AixLib you should always use 2 elements!!!` -> what should be written there?

See also [PR 1546 in AixLib](https://github.com/RWTH-EBC/AixLib/pull/1546)

Looking forward to your feedback on the changes!


PGorzalka:
@DaJansenGit added your commits to the CI branch and renamed the ground temperature selector in the records as you suggested in the AixLib PR

teaser/examples/e11_include_adjacent_zones_and_varying_soil_temps.py
```diff
@@ -0,0 +1,647 @@
...
+import shutil
+
+
+def example_interzonal_single_building():
+    """
+    This example features the data presented in a contribution to the 2023
+    International Modelica Conference: https://doi.org/10.3384/ecp204577
+    """
+    # for the conference publication, four variations were calculated
+    for variation in ['A', 'B', 'C', 'D']:
```
DaJansenGit, line 43: Maybe add more information about what these variants are. You could add the table from the paper?

teaser/examples/examplefiles/ASHRAE140_600.json
```diff
@@ -13,7 +13,7 @@
             "current_solver": "Cvode",
             "equidistant_output": true,
             "results_at_events": false,
-            "version": "3.2.2"
+            "version": "4.0.0"
```
DaJansenGit, line 16: nice catch, that has been there for a while :D

teaser/data/output/modelicatemplate/AixLib/AixLib_Multizone
```diff
@@ -20,7 +26,7 @@ AixLib.ThermalZones.ReducedOrder.Multizone.MultizoneEquipped multizone(
     % else:
         use_C_flow = false,
         use_moisture_balance = false,
-        redeclare package Medium = Modelica.Media.Air.SimpleAir,
+        redeclare package Medium = ${"Modelica.Media.Air.DryAirNasa" if any([tz.use_conditions.with_heating is False for tz in bldg.thermal_zones]) else "Modelica.Media.Air.SimpleAir"},
```
DaJansenGit, line 29: This is a nice change. Should also prevent model crashes in the future with normal exports.

teaser/data/input/teaserjson_input.py
```diff
@@ -269,6 +297,42 @@ def load_teaser_json(path, project):
                 ceil.name = cl_name
                 set_basic_data_teaser(cl_in, ceil)
                 set_layer_data_teaser(cl_in, ceil)
+            try:
```
DaJansenGit, line 300: I have to admit that I do not really like the the try/except approach here. I would suggest to use `get` functionality to retrieve default values if the key could not be found. What do you think about this? If you agree please change this also at the other occurenes in this function

teaser/logic/buildingobjects/buildingphysics/buildingelement.py
```diff
@@ -309,10 +329,26 @@ def load_type_element(
...
+                )
+            except KeyError:
+                warnings.warn(
+                    ('Type element ' + type_element_key + ' was not found. '
+                     + 'Going back to default element for year and construction'
+                     + '...')
+                )
+                type_element_key = None
+
+        if not type_element_key:
```
DaJansenGit, line 346: I couldn't find a point in the code where you are actually using the `type_element_key`/`key_str` argument. Am I missing something or is this a relict and can be removed?

If not, I would suggest to use the same name `type_element_key` for both arguments.

teaser/data/input/buildingelement_input_json.py
```diff
@@ -60,6 +76,51 @@ def load_type_element(element, year, construction, data_class):
                     f"year{year}, element={type(element).__name__}")
 
 
+def load_type_element_by_key(element, key_str, data_class,
```
DaJansenGit, line 79: I couldn't find a point in the code where you are actually using the type_element_key/key_str argument. Am I missing something or is this a relict and can be removed?

If not, I would suggest to use the same name type_element_key for both arguments.

(Left the same commend in buildingelement.py)

DaJansenGit (changes requested):
Thanks for this huge adjustments and new features in TEASER. I have some minor remarks and I recognized that the CI failed. But only for python 3.12.

Furthermore, I would like to check if existing results for all examples all stay the same as this is currently not done via the CI in TEASER. 

Also I have to say that I didn't check if the implementation for the R/C calculation are correctly as these are a lot of changes. If the existing models all pass together with your paper I guess this is fine.

teaser/examples/e11_include_adjacent_zones_and_varying_soil_temps.py
PGorzalka: I added a reference to the Table. As the paper is openly available, I think that should be fine.

teaser/logic/buildingobjects/buildingphysics/buildingelement.py
PGorzalka: Indeed, it has vanished from the code and also from our paper. In fact, I needed it for the additional variations E and F that I calculated for my dissertation (which has finally been published :D see Table 4.2 [here](https://publications.rwth-aachen.de/record/998671)). I consider it's a nice feature if you want to load some element that does not match the current archetype. 

So I implemented your second suggestion. I don't know if something was changed in the meantime, but using this needs to have year and construction being keyword arguments. Hope it's fine like I did it now.
As it's not tested at the moment, I also added it to the example to make sure it's not "forgotten" in the future.

teaser/data/input/buildingelement_input_json.py
PGorzalka: done

PGorzalka:
Regarding the CI failing on Python 3.12, does this only happen for this PR or generally?

PGorzalka:
@DaJansenGit is there anything I can do here?

teaser/logic/buildingobjects/buildingphysics/buildingelement.py
DaJansenGit: Thank you for the changes. The solution is fine for me.

teaser/examples/e11_include_adjacent_zones_and_varying_soil_temps.py
DaJansenGit: Thanks for adding the changes and happy that I could help :)

teaser/data/input/teaserjson_input.py
DaJansenGit: Thanks for adding the changes and I'm happy that I could help :)

DaJansenGit:
Hey, sorry it's taken so long, I've been down a rabbit hole for the last few weeks writing my thesis.

I looked it up and this seems to be a common problem due to the new floating point calculation in Python 3.12. You're just the unfortunate person doing the first PR since we implemented the tests for 3.12. :D I'll make a fix in the main branch, add it to this branch and then we can proceed.

PGorzalka:
Thank you @DaJansenGit  :D 
Hope your thesis is going well!
But I cannot take that honor, privileges are missing ;)

PR #788, opened 2024-11-27
787 dynamic heating and cooling profiles

### LINKED ISSUE #787: dynamic heating and cooling profiles
DIN EN 18599-10 offers dynamic profiles for 24 h with night setback for heating and cooling set temperatures. Currently, TEASER just uses static values. This will be changed to the dynamic values from DIN EN 18599-10.

As the current values for the PI-controlers would lead to very high peaks in heating demands, we will adjust those as well.

DaJansenGit:
closes #787


FWuellhorst (changes requested):
The changes look good, but I think you have to increase the version number to 1.1.0, as the change may have a large impact on the results. Are the setbacks now the new default for archetype export? If so, maybe a quick comparison of how heat demands change would be good for release notes, what do you think?

DaJansenGit:
> The changes look good, but I think you have to increase the version number to 1.1.0, as the change may have a large impact on the results. Are the setbacks now the new default for archetype export? If so, maybe a quick comparison of how heat demands change would be good for release notes, what do you think?

I agree. I think it makes sense to use those as new defaults but you are right, we should at least give some information what the effects of these changes are. I will provide some wrap up here about how the results change for the examples before we merge this and will increase version to `v1.1.0`. Thanks :)

HvanderStok:
Hi, I noticed that the profiles for "Living" are not changed, but these are used for the residential buildings. Can you change them as well?

DaJansenGit:
> Hi, I noticed that the profiles for "Living" are not changed, but these are used for the residential buildings. Can you change them as well?

Hey, those were not adjusted because DIN EN 18599-10 is only for non residential buildings and doesn't offer data for living. The data for `Living` is coming from SIA 2024 which does not offer temperature profiles. But we can just add some night set back. Actually when looking at SIA 2024 I found that they offer a Living-MFH and Living-SFH dataset. It seems like the current data is based on the Living-SFH data, so I will leave this data set as it is (apart from the heating and cooling profiles) to stay backwards compatible and add Living-MFH as well.

DaJansenGit:
Some documentation regarding the changes:
* I assume the current values for Living are taken from/based on SIA 2024 for Living-SFH based on the person related internal heat gains
* Living-SFH and Living-MFH only difference is the persons/m². For SFH its 50m²/person and MFH 30m²/person. I think it makes sense to implement the MFH version here, as this makes an actual difference in the internal loads and it seems valid to me that in SFHs do generally have more space per person than MFHs
* Machine internal gains are not fitting the data from SIA 2024 (existing data says for existing Living data but I won't change this 2 W/m², but SIA 2024 would suggest 8 W/m². Looking at DIN V 4108-6, Table 2 suggests 5 W/m² as an average value per day. But from my personal feeling both 5 W/m² and 8 W/m² feels a bit high. I will leave the value of 2 W/m² for both (SFH and MFH)
* Profiles for persons and machines are the same in SIA 2024 for EFH and MFH, but do not fit the existing data in TEASER. I actually have no idea where the existing data is taken from... T.b.d. if we change this. The profiles in SIA have less appearance of persons in general and less appearance of machine gains. Changing the data profile data to SIA values would therefore decrease the internal loads in the simulation and thus increase simulated heating demand.

DaJansenGit:
# Changes to simulation results
* Generated with `e2_export_aixlib_models.py`,
* for benchmark with `main` branch on commit c048339 
* for the new usage conditions with commit 9f11638
## Living
image
## Office
image
## Institute
image

FWuellhorst:
@DaJansenGit The values in the paper are an average for the whole German building stock, simulated by Michael. I also think 30-40 % is a bit much. Maybe @HvanderStok can shed more light on possible savings when comparing the calculations to demand data and BESMod simulations. I could image that the walls get really cold as the ideal heater only heats the volume, but otherwise the trajectories look ok.

PR #796, opened 2024-12-18
795 besmod examples without hydraulic

### LINKED ISSUE #795: BESMod examples without hydraulic
When exporting BESMod examples without a hydraulic system, not specifying `THydSup_nominal` causes an error, but it is not needed for these systems.

HvanderStok:
Closes #795


FWuellhorst (approved):
Looks good!

PR #799, opened 2025-01-10
make t_bt customizable

### LINKED ISSUE #798: Make t_bt a parameter
The frequency parameter `t_bt` influences the capacities. While well set according to VDI (see https://publications.rwth-aachen.de/record/749705/files/749705.pdf), different values may be interesting to obtain to analyze different building characteristics.

FWuellhorst:
Closes #798 
Intentionally does not make it a parameter in building and just an attribute to avoid users changing it "by accident".

### DIFF
--- a/teaser/examples/e6_generate_building.py
+++ b/teaser/examples/e6_generate_building.py
@@ -258,6 +258,7 @@ def example_create_building():
         win_material.name = "GlasWindow"
         win_material.thermal_conduc = 0.067
         win_material.transmittance = 0.9
+    return prj
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


FWuellhorst:
To document, the specific heat capacity changes accordingly when changing t_bt for values betwenn 1 h and 7 d, typical values reported in https://publications.rwth-aachen.de/record/749705/files/749705.pdf:
image

DaJansenGit (approved):
Thanks for addressing this issue. I think this not only makes the code better but also improves the comprehensibility regarding the `t_bt` attribute.

PR #809, opened 2025-01-28
Fix custom paths

### LINKED ISSUE #808: Bug in custom export path
The code fails when exporting a project with a custom path. Currently, the custom path is appended to the default teaser path, resulting in the error.

FelixStege:
Closes #808


DaJansenGit (dismissed):
Thanks for the changes!

DaJansenGit (approved):

PR #817, opened 2025-04-02
816 minor bug fixes

### LINKED ISSUE #816: Minor Bug fixes
There are two minor bugs:
1. When exporting a project, it is not checked if it includes any buildings, which causes an unclear error message.
2. A wrong warnings call for a large base infiltration.

HvanderStok:
Closes #816.

### DIFF
--- a/teaser/project.py
+++ b/teaser/project.py
@@ -771,6 +771,15 @@ class Project(object):
 
         tjson_in.load_teaser_json(path, self)
 
+    def _check_buildings(self, internal_id: Optional[float] = None):
+        """Utility method to verify that the project has buildings and,
+        if an internal_id is provided, that it exists.
+        """
+        if not self.buildings:
+            raise ValueError("The project includes no buildings to export.")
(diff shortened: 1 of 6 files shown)


teaser/project.py
```diff
@@ -981,6 +991,11 @@ def export_ibpsa(self, library="AixLib", internal_id=None, path=None):
             "IDEAS",
         ], ass_error_1
 
+        if not self.buildings:
+            raise ValueError("The project includes no buildings to export.")
+        if internal_id is not None and not any(bldg.internal_id == internal_id for bldg in self.buildings):
+            raise ValueError(f"Building with internal_id {internal_id} not found in the project.")
+
```
FWuellhorst, line 998: maybe add a util function for this check, as it is always the same? Else, looks good.

FWuellhorst (dismissed):
Thanks for the fix, pls consider my comment

FWuellhorst (approved):

